PR #6, opened 2019-07-03
Vladimir's Solution

VladimirTi:


src/components/Buttons.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+function Buttons({ onClickHandler, isRequested }) {
```
YuriiHoliuk: `Button`, or as it is not general button component but specific only for request sending can be named `LoadDataButton` etc.

src/components/Comment.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+function Comment({ comment }) {
```
YuriiHoliuk: Why do you need to pass the whole object as a prop? will be better to pass name and body separately

src/components/MainHandler.js
```diff
@@ -0,0 +1,19 @@
...
+function mapStateToProps(state) {
+  return {
+    dataRequested: state.dataRequested,
+    postsList: state.postsList
+  };
+}
+
+function mapDispatchToProps(dispatch) {
+  return {
+    request: () => dispatch(dataRequested()),
```
YuriiHoliuk: Can be used short version(as object) here

src/redux/actions.js
```diff
@@ -0,0 +1,41 @@
...
+      type: DATA_REQUESTED
+    });
+    Promise.all([
+      getData('https://jsonplaceholder.typicode.com/posts'),
+      getData('https://jsonplaceholder.typicode.com/users'),
+      getData('https://jsonplaceholder.typicode.com/comments')
+    ])
+      .then(([posts, users, comments]) => {
+        const commentsMap = {};
+        comments.forEach(el => {
```
YuriiHoliuk: так собі виглядає, краще б reduce

PR #8, opened 2019-07-16
Solution

SherJlok15:


src/components/Comment.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+function Comment(props) {
+  return (
+    <div className="comment">
+      <span className={props.search ? "post-cross-none" :"comment-cross"} onClick={() => props.removeComment(props.index, props.posts, props.postIndex)}>&times;</span>
```
gestap: do not write all in one line

gestap: you use `props.search` but you didn't add it to `mapStateToProps`

src/components/PostList.js
```diff
@@ -0,0 +1,36 @@
...
+        <div className="lds-dual-ring"></div>
+      </div>
+    );
+  }
+  return (
+    <>
+    <input type="text" placeholder="Search" onChange={(event) => props.getInputValue(event)}/>
+    <div className="posts">
+      {
+        props.search ?
```
gestap: format the code
https://github.com/mate-academy/style-guides/blob/master/javascript.md#comparison-operators--equality

PR #61, opened 2020-05-30
Fix

OlEzhA0:
[LINK](https://olezha0.github.io/react_redux-list-of-posts/)


src/store/index.ts
```diff
@@ -1,57 +1,25 @@
...
-    } catch (error) {
-      dispatch(setMessage('Error occurred when loading data'));
-    }
-
-    dispatch(finishLoading());
-  };
-};
+export const getPosts = (state: RootState) => state.posts;
+export const isLoaded = (state: RootState) => state.loaded;
+export const queryValue = (state: RootState) => state.query;
```
irenhh, line 19: there is no need to store query for some input in the global store

src/store/loaded.ts
```diff
@@ -0,0 +1,17 @@
+import { AnyAction } from 'redux';
+
+const LOADED = 'LOADED';
```
irenhh, line 3: is loaded?

src/store/loading.ts
```diff
@@ -1,24 +1,17 @@
...
 
-const reducer = (loading = false, action: AnyAction) => {
+const loadingReducer = (state = false, action: AnyAction) => {
   switch (action.type) {
-    case START_LOADING:
-      return true;
-
-    case FINISH_LOADING:
-      return false;
+    case LOADING:
```
irenhh, line 9: the same with naming

src/store/index.ts
```diff
@@ -1,57 +1,22 @@
...
-/**
- * Each concrete reducer will receive all the actions but only its part of the state
- *
- * const rootReducer = (state = {}, action) => ({
- *   loading: loadingReducer(state.loading, action),
- *   message: messageReducer(state.message, action),
- * })
- */
 const rootReducer = combineReducers({
   loading: loadingReducer,
```
irenhh, line 7: fields must have names like `isLoading` as well if they store bool values

PR #108, opened 2022-03-17
Develop

mykola-koval:
[DEMO LINK](https://mykola-koval.github.io/react_redux-list-of-posts/)


Olexandr6 (approved):
Would be better to add loader when you're fetching data from server

PR #131, opened 2022-06-28
solution

vlad-demchuk:
[DEMO LINK](https://vlad-demchuk.github.io/react_redux-list-of-posts/)


src/App.tsx
```diff
@@ -1,20 +1,92 @@
...
+              <option value={user.id} key={user.id}>{user.name}</option>
+            ))}
+          </select>
+        </label>
+      </header>
+
+      <main className="App__main">
+        <div className="App__sidebar">
+          <PostsList
+            posts={posts}
```
SavMary, line 79: You can get posts from redux directly inside PostList

vlad-demchuk: I'm not sure, I have filter by users here

Denchepornyuk (approved):
good job 👍

PR #142, opened 2022-07-28
Test commit

vsupruniuk:
[DEMO LINK](https://vsupruniuk.github.io/react_redux-list-of-posts/)


src/store/users.ts
```diff
@@ -0,0 +1,59 @@
...
+  setSelectedUserId: (userId: number) => {
+    return { type: SET_SELECTED_USER_ID, userId };
+  },
+};
+
+const usersReducer = (state = initialState, action: AnyAction) => {
+  switch (action.type) {
+    case SET_USERS:
+      return {
+        ...action,
```
Olexandr6, line 38: Why here is `...action` not `...state`?

src/store/users.ts
```diff
@@ -0,0 +1,59 @@
...
+    case SET_LOAD_USERS_ERROR:
+      return {
+        ...state,
+        loadUsersError: true,
+      };
+
+    case SET_SELECTED_USER_ID:
+      return {
+        ...state,
+        selectedUserId: action.selectedUserId,
```
Olexandr6, line 51: In the `actions.setSelectedUserId` above, you return `userId`, but here you are trying to set `selectedUserId`:
image

Olexandr6 (changes requested):

src/App.tsx
```diff
@@ -1,20 +1,86 @@
...
+          {loadPostsError ? (
+            <h1>Failed to load posts</h1>
+          ) : (
+            <PostsList
+              posts={posts}
+              selectedPostId={selectedPostId}
+              handleSelectedPostId={handleSelectedPostId}
+              reloadPosts={() => dispatch(loadPosts(selectedUserId))}
+            />
+          )}
```
Olexandr6, line 75: Would be great to add `Loader`

Olexandr6 (approved):
Let's fix this case:
image

PR #166, opened 2022-10-10
first commit

pavel-gutsal:
first commit
new CommentForm wasn`t touched
https://pavel-gutsal.github.io/react_redux-list-of-posts/


src/components/PostDetails.tsx
```diff
@@ -1,92 +1,51 @@
...
-    // we delete the comment immediately so as
-    // not to make the user wait long for the actual deletion
-    setComments(
-      currentComments => currentComments.filter(
-        comment => comment.id !== commentId,
-      ),
-    );
+    dispatch(removeComment(commentId));
 
     await commentsApi.deleteComment(commentId);
```
polosanya, line 50: You are not handling possible errors in your async functions. What if your request will be failed?
You can use `try...catch` to avoid those situations. Please fix all cases
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Statements/try...catch

src/components/UserSelector.tsx
```diff
@@ -69,20 +64,25 @@ export const UserSelector: React.FC<Props> = ({
...
-          ))}
+          {
+            users && users.length > 0 && (
+              users.map(user => (
+                <a
+                  key={user.id}
+                  href={`#user-${user.id}`}
+                  onClick={() => {
+                    dispatch(changeAuthor(user));
+                    // onChange(user);
```
polosanya, line 75: Don't forget to remove useless comments

src/features/comments/comments.ts
```diff
@@ -0,0 +1,85 @@
+/* eslint-disable no-console */
+/* eslint-disable no-param-reassign */
```
polosanya, line 2: Disabling eslint is a bad pratice. Try to fix problems that trigger it.

polosanya (changes requested):
Nice start, please check my comments)
Btw, why didn't you delete default redux example on top of the page?
image

PR #169, opened 2022-10-26
Develop

YaroslavKolbaiev:
[DEMO LINK](https://YaroslavKolbaiev.github.io/react_redux-list-of-posts/)


src/api/comments.ts
```diff
@@ -1,8 +1,7 @@
-// import { client } from '../utils/axiosClient';
 import { client } from '../utils/fetchClient';
 import { Comment } from '../types/Comment';
 
-export const getPostComments = (postId: number) => {
+export const getPostComments = (postId: number | undefined) => {
```
qoa11a, line 4: Better check if value is undefined on upper level and just don't call a function if it is.

qoa11a (approved):

PR #203, opened 2023-01-26
Dome

Nazar-Brunarskyi:
[DEMO LINK](https://Nazar-Brunarskyi.github.io/react_redux-list-of-posts/)


src/App.tsx
```diff
@@ -53,7 +55,7 @@ export const App: React.FC = () => {
           <div className="tile is-parent">
             <div className="tile is-child box is-success">
               <div className="block">
-                <UserSelector value={author} onChange={setAuthor} />
+                <UserSelector value={author} onChange={onSetAuthor} />
```
Oleksandr-Filo, line 58: I think you don't need to pass props, you can get state from redux inside the component

Oleksandr-Filo (approved):
looks good, just one thing

valerii-derkach (approved):
🔥

PR #207, opened 2023-02-02
Done

ruslik87:
[DEMO LINK](https://ruslik87.github.io/react_redux-list-of-posts/)


SanyaBratashchuk (changes requested):
Fix route 
image

src/App.tsx
```diff
@@ -76,18 +70,14 @@ export const App: React.FC = () => {
                   </div>
                 )}
 
-                {author && loaded && !hasError && posts.length === 0 && (
+                {author && !loaded && !hasError && posts.length === 0 && (
```
loralevitska, line 73: Long expressions (more than 2 operand) move to variable. Fix all cases please

src/components/PostDetails.tsx
```diff
@@ -1,126 +1,64 @@
...
+  const handleDeleteComment = (commentId: number) => {
+    dispatch(deleteComments(commentId));
   };
 
   return (
     <div className="content" data-cy="PostDetails">
       <div className="block">
         <h2 data-cy="PostTitle">
-          {`#${post.id}: ${post.title}`}
+          {`#${selectedPost?.id}: ${selectedPost?.title}`}
```
loralevitska, line 36: Better to check if there is a selected post for the entire block (and then render if there is) than to check for each variable

src/components/PostsList.tsx
```diff
@@ -1,56 +1,58 @@
...
-    <table className="table is-fullwidth is-striped is-hoverable is-narrow">
-      <thead>
-        <tr className="has-background-link-light">
-          <th>#</th>
-          <th>Title</th>
-          <th> </th>
-        </tr>
-      </thead>
+  return (
+    <div data-cy="PostsList">
```
loralevitska, line 17: Use semantic tags for list `ul` `li`. Don't forget to remove default li dots

src/features/authorSlice.ts
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable no-param-reassign */
```
loralevitska, line 1: Try to fix eslint error and remove comment. Fix everywhere

loralevitska (changes requested):
Good job! There are things that should be fixed:

 I can add empty comment. PLease fix it (use `string.trim()`)
image

SanyaBratashchuk (approved):

PR #209, opened 2023-02-05
Implement requirements

UmizDemud:
[DEMO LINK](https://UmizDemud.github.io/react_redux-list-of-posts/)


src/App.tsx
```diff
@@ -86,7 +95,7 @@ export const App: React.FC = () => {
                   <PostsList
                     posts={posts}
                     selectedPostId={selectedPost?.id}
-                    onPostSelected={setSelectedPost}
+                    onPostSelected={selectPost}
```
ysrckr, line 98: You have these in the store. Can't you just pull it from state in that component instead of passing it as a prop?

ysrckr (changes requested):
Code looks good but the demo link doesn't work. It would be great if you could check your demo link.

ysrckr (approved):
I think it is because of github pages

src/features/selectedPost/selectedPostSlice.ts
```diff
@@ -0,0 +1,32 @@
...
+  reducers: {
+    setSelectedPost: (state, action: PayloadAction<Post | null>) => {
+      state.value = action.payload;
+    },
+  },
+});
+
+export const { setSelectedPost } = selectedPostSlice.actions;
+
+export const selectSelectedPost = (
```
mgrinko, line 28: With typescript I don't think you need selectors, because you can just select what you need in the component

src/features/users/usersSlice.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const loadUsers = createAsyncThunk(
+  'users/getUsers',
+  async () => {
+    const value = await getUsers();
+
+    // The value we return becomes the `fulfilled` action payload
+    return value;
+  },
+);
```
mgrinko, line 26: can be shortened to
```suggestion
export const loadUsers = createAsyncThunk('users/getUsers', getUsers);
```

src/features/comments/commentsSlice.ts
```diff
@@ -0,0 +1,98 @@
...
+};
+
+export const loadComments = createAsyncThunk(
+  'comments/getPostComments',
+  async (postId: number) => {
+    const value = await getPostComments(postId);
+
+    // The value we return becomes the `fulfilled` action payload
+    return value;
+  },
```
mgrinko, line 25:
```suggestion
  (postId: number) => getPostComments(postId),
```

src/features/comments/commentsSlice.ts
```diff
@@ -0,0 +1,98 @@
...
+    return commentId;
+  },
+);
+
+export const commentsSlice = createSlice({
+  name: 'comments',
+  initialState,
+  reducers: {
+    clear: (state) => {
+      state.value = [];
```
mgrinko, line 53: `state.value` is not really a meaningful name

src/features/comments/commentsSlice.ts
```diff
@@ -0,0 +1,98 @@
...
+      })
+      .addCase(removeComment.pending, (state) => {
+        state.status = 'loading';
+      })
+      .addCase(removeComment.fulfilled, (state) => {
+        state.status = 'items';
+      })
+      .addCase(removeComment.rejected, (state) => {
+        state.status = 'hasError';
+      });
```
mgrinko, line 91: I think you should check RTK Query to simplify this

src/components/PostDetails.tsx
```diff
@@ -2,94 +2,42 @@ import React, { useEffect, useState } from 'react';
...
-        comment => comment.id !== commentId,
-      ),
-    );
 
-    await commentsApi.deleteComment(commentId);
-  };
+    setLoaded(false);
+    dispatch(loadComments(post.id))
+      .then(postsFromServer => {
+        setError(postsFromServer.meta.requestStatus === 'rejected');
```
mgrinko, line 33: I think you should change you loadComments method so it rejects on error

mgrinko (approved):
Take my comments into account

PR #211, opened 2023-02-10
Initial commit

artem-hirzhev:
[DEMO LINK](https://artem-hirzhev.github.io/react_redux-list-of-posts/)


SanyaBratashchuk (changes requested):
Remove it 
image

src/components/PostDetails.tsx
```diff
@@ -65,28 +34,14 @@ export const PostDetails: React.FC<Props> = ({ post }) => {
...
 
-      setComments(
-        currentComments => [...currentComments, newComment],
-      );
-
-      // setComments([...comments, newComment]);
-      // works wrong if we wrap `addComment` with `useCallback`
-      // because it takes the `comments` cached during the first render
-      // not the actual ones
+      dispatch(setComments([...comments, newComment]));
```
Volodymyr-Melnyk-Fron10der, line 37: Looks like we should create addComment action to add new comment
```suggestion
      dispatch(addComment(newComment));
```

src/components/PostDetails.tsx
```diff
@@ -65,28 +34,14 @@ export const PostDetails: React.FC<Props> = ({ post }) => {
...
 
   const deleteComment = async (commentId: number) => {
-    // we delete the comment immediately so as
-    // not to make the user wait long for the actual deletion
-    setComments(
-      currentComments => currentComments.filter(
-        comment => comment.id !== commentId,
-      ),
-    );
+    dispatch(setComments(comments.filter(comment => comment.id !== commentId)));
```
Volodymyr-Melnyk-Fron10der, line 44: Same here. 
If we work with redux let's use all it's power. Move this logic inside commentsSlice and create a new action with will only delete comments by id
```suggestion
    dispatch(deleteComment(commentId));
```

src/components/PostsList.tsx
```diff
@@ -1,56 +1,53 @@
...
-export const PostsList: React.FC<Props> = ({
-  posts,
-  selectedPostId = 0,
-  onPostSelected,
-}) => (
-  <div data-cy="PostsList">
-    <p className="title">Posts:</p>
+export const PostsList: React.FC = () => {
+  const posts = useAppSelector(state => state.posts.items);
+  const selectedPost = useAppSelector(state => state.selectedPost.selectedPost);
```
Volodymyr-Melnyk-Fron10der, line 8: 1. I suggest to create and export all selectors callbacks from certain slice....ts file
2. It looks a little strange `selectedPost.selectedPost`, try to avid embeddedness like this

```suggestion
// postsSlice.ts
... 
 export const selectPosts = state => state.posts.items;
 export const selectSelectedPost = state => state.selectedPost.selectedPost;
 
export const { set } = postsSlice.actions;
export default postsSlice.reducer;

 // component.tsx
 ...
  const posts = useAppSelector(selectPosts);
  const selectedPost = useAppSelector(selectSelectedPost );
...
```

src/components/PostsList.tsx
```diff
@@ -1,56 +1,53 @@
...
+                  className={classNames(
+                    'button',
+                    'is-link',
+                    {
+                      'is-light': post.id !== selectedPost?.id,
+                    },
+                  )}
+                  onClick={() => {
+                    dispatch(set(post.id === selectedPost?.id ? null : post));
+                  }}
```
Volodymyr-Melnyk-Fron10der, line 42: 1. Move this callback to a certain function
2. `set` what? Provide more meaningful name

src/components/PostsList.tsx
```diff
@@ -1,56 +1,53 @@
...
+                >
+                  {post.id === selectedPost?.id ? 'Close' : 'Open'}
```
Volodymyr-Melnyk-Fron10der, line 44: You repeat `post.id === selectedPost?.id `condition a several times. So, maybe it`s better to move it to certain variable and reuse it ?

src/components/UserSelector.tsx
```diff
@@ -74,7 +63,7 @@ export const UserSelector: React.FC<Props> = ({
               key={user.id}
               href={`#user-${user.id}`}
               onClick={() => {
-                onChange(user);
+                dispatch(setSelectedUser(user));
               }}
```
Volodymyr-Melnyk-Fron10der, line 67: Might be, but I suggest to move callback into separated function if you need to write it with more than 1 line

Volodymyr-Melnyk-Fron10der (changes requested):
Almost done! But there are some issues you could improve:

src/components/PostsList.tsx
```diff
@@ -1,56 +1,58 @@
...
+                      'is-light': post.id !== selectedPost?.id,
+                    },
+                  )}
+                  onClick={() => handleClick(post)}
+                >
+                  {post.id === selectedPost?.id ? 'Close' : 'Open'}
+                </button>
+              </td>
+            </tr>
+          ))}
```
Volodymyr-Melnyk-Fron10der, line 53: That's what I meant
```suggestion
          {posts.map(post => {
          const isOpenedPost = post.id === selectedPost?.id;
          
          return (
            <tr key={post.id} data-cy="Post">
              <td data-cy="PostId">{post.id}</td>
              <td data-cy="PostTitle">{post.title}</td>
              <td className="has-text-right is-vcentered">
                <button
                  type="button"
                  data-cy="PostButton"
                  className={classNames(
                    'button',
                    'is-link',
                    {
                      'is-light': !isOpenedPost ,
                    },
                  )}
                  onClick={() => handleClick(isOpenedPost ? null : post)}
                >
                  {isOpenedPost  ? 'Close' : 'Open'}
                </button>
              </td>
            </tr>
          )})}
```

Volodymyr-Melnyk-Fron10der (changes requested):
Great, but one last comment there

PR #212, opened 2023-02-10
Solution_1

Lenur29:
https://Lenur29.github.io/react_redux-list-of-posts/


Anastasiia-Svintsova (changes requested):
- Hey! Why do we need this part? 
image

- Check the error handling. The existing comments shouldn't disappear. And I can't open comments when the internet connection is good again. Fix it, please 

https://user-images.githubusercontent.com/99403198/218255942-4bec04d6-9399-4a05-a859-a88495809382.mov

src/App.tsx
```diff
@@ -8,52 +8,38 @@ import { PostsList } from './components/PostsList';
...
-      // We disable the spinner in any case
-      .finally(() => setLoaded(true));
-  }
+  const {
+    posts,
+    loaded,
+    hasError,
+  } = useAppSelector(state => state.posts);
+  const selectedPost = useAppSelector(state => state.selectedPost);
+  const author = useAppSelector(state => state.author);
```
Volodymyr-Melnyk-Fron10der, line 22: I suggest to move all selectors callbacks to the `slice....ts file` and export them from there
```

export selectPost = state => state.selectedPost
...
const selectedPost = useAppSelector(selectPost);
```

src/features/posts/selectedPostSlice.ts
```diff
@@ -0,0 +1,20 @@
...
+const initialState: PostState = null;
+
+export const selectedPostSlice = createSlice({
+  name: 'selectedPost',
+  initialState: initialState as PostState,
+  reducers: {
+    selectPost: (_state, action: PayloadAction<Post | null>) => {
+      return action.payload;
+    },
+  },
```
Volodymyr-Melnyk-Fron10der, line 15:
```suggestion
  reducers: {
    selectPost: (_state, action: PayloadAction<Post | null>) => {
      return action.payload;
    },
    
    resetPost: () => initialState;
  },
```

src/App.tsx
```diff
@@ -8,52 +8,38 @@ import { PostsList } from './components/PostsList';
...
+  const dispatch = useAppDispatch();
 
   useEffect(() => {
-    // we clear the post when an author is changed
-    // not to confuse the user
-    setSelectedPost(null);
+    dispatch(selectPost(null));
```
Volodymyr-Melnyk-Fron10der, line 26: why name is selectPost, but actually you are removing the post?

*check my suggestion in `postsSlice.ts`
```suggestion
    dispatch(resetPost());
```

src/components/PostsList.tsx
```diff
@@ -1,56 +1,61 @@
...
           </tr>
-        ))}
-      </tbody>
-    </table>
-  </div>
-);
+        </thead>
+
+        <tbody>
+          {posts.map(post => (
```
Volodymyr-Melnyk-Fron10der, line 31: Decomposing again 
```suggestion
          {posts.map(({ id, title, .... }) => (
```

src/components/PostsList.tsx
```diff
@@ -1,56 +1,61 @@
...
+                      'is-light': post.id !== selectedPost?.id,
+                    },
+                  )}
+                  onClick={() => {
+                    dispatch(selectPost(post.id === selectedPost?.id
+                      ? null
+                      : post));
+                  }}
+                >
+                  {post.id === selectedPost?.id ? 'Close' : 'Open'}
```
Volodymyr-Melnyk-Fron10der, line 52: Keep DRY(don't repeat yourself) pattern.
You have created condition `post.id === selectedPost?.id` 3 times, so move it to a variable and use it instead

src/features/comments/commentsSlice.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export const commentsSlice = createSlice({
+  name: 'comments',
+  initialState,
+  reducers: {
+    removeComment: (state, action: PayloadAction<number>) => {
+      state.comments = state.comments.filter(comment => (
+        comment.id !== action.payload
+      ));
+    },
```
Volodymyr-Melnyk-Fron10der, line 53: **(Optional)** I would do like this
```suggestion
    removeComment: (state, { payload: commentId }: PayloadAction<number>) => {
      state.comments = state.comments.filter(comment => (
        comment.id !== commentId
      ));
    },
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi!  Good job, move on👍

1. There is no error if deleting was failed
image

src/App.tsx
Lenur29: Hi, thank you for the correction) I've checked this working demo https://mate-academy.github.io/react_redux-list-of-posts/ and it seems to be that we do not handle the deleting process

Anastasiia-Svintsova (approved):

PR #219, opened 2023-02-13
Implement list of posts using redux

anastasiia-tilikina:
[DEMO LINK](https://anastasiia-tilikina.github.io/react_redux-list-of-posts/)

I used RTK Query so I omitted the `wait` function since I haven't made any promises or fetches manually.
Also, I haven't discarded the default cache behavior. So once users' posts or comments were downloaded, the app won't show the loader on the next click on the same user or post.
Everything else works as desired.


AlisaVlasova (approved):
:fire:

PR #220, opened 2023-02-15
solution

Lun4trick:
[DEMO](https://lun4trick.github.io/react_redux-list-of-posts/)


src/components/NewCommentForm.tsx
```diff
@@ -57,15 +60,14 @@ export const NewCommentForm: React.FC<Props> = ({ onSubmit }) => {
...
 
-    setSubmitting(true);
-
-    // it is very easy to forget about `await` keyword
-    await onSubmit({ name, email, body });
-
-    // and the spinner will disappear immediately
-    setSubmitting(false);
+    dispatch(addComment({
+      id: 0,
```
nurdanokcu, line 64: Instead of assigning id as 0 hero you can change the type in commentSlice for the addComment function so that you wouldn't need to add id here.

nurdanokcu (approved):
Nice.

src/features/CommentSlice.ts
```diff
@@ -0,0 +1,83 @@
+/* eslint-disable no-param-reassign */
+import { createAsyncThunk, createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { getPostComments, createComment, deleteComment } from '../api/comments';
+// eslint-disable-next-line import/no-cycle
```
meforoma, line 4: better avoid circular imports
move the exported entity into a separate file and import into other files from there

meforoma (approved):
decent ✅

PR #228, opened 2023-03-02
solution

savchuksavchuk:
[DEMO LINK](https://savchuksavchuk.github.io/react_redux-list-of-posts/)


src/components/PostDetails.tsx
```diff
@@ -1,96 +1,43 @@
...
-    loadComments();
-  }, []);
+    dispatch(loadComments(post.id));
+    setVisible(false);
+  }, [post.id]);
 
-  useEffect(loadComments, [post.id]); // Wrong!
-  // effect can return only a function but not a Promise
-  */
+  const handlerDeleteComment = (commentId: number) => {
```
OleziO, line 30: Rename to `handleDeleteComment`, its better name.

OleziO (changes requested):
image
 Validate Email address

src/app/store.ts
```diff
@@ -1,10 +1,18 @@
 import { configureStore, ThunkAction, Action } from '@reduxjs/toolkit';
 // eslint-disable-next-line import/no-cycle
-import counterReducer from '../features/counter/counterSlice';
+import usersReducer from '../features/counter/usersSlice';
+import authorReducer from '../features/counter/authorSlice';
+import postsReducer from '../features/counter/postsSlice';
+import selectedPostReducer from '../features/counter/selectedPostSlice';
```
anstsot, line 6: it would be great to place the slices directly in `features` folder

src/features/counter/postsSlice.ts
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable no-param-reassign */
+import { createAsyncThunk, createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { getUserPosts } from '../../api/posts';
+import { Post } from '../../types/Post';
+
+export const loadUsersPosts
+  = createAsyncThunk('posts/fetch', (authorId: number) => {
+    return getUserPosts(authorId);
+  });
```
anstsot, line 9: just small improvement for better readability:
```suggestion
export const loadUsersPosts = createAsyncThunk(
  'posts/fetch',
  (authorId: number) => getUserPosts(authorId),
);
```

src/components/NewCommentForm.tsx
```diff
@@ -44,6 +44,14 @@ export const NewCommentForm: React.FC<Props> = ({ onSubmit }) => {
     setErrors(current => ({ ...current, [field]: false }));
   };
 
+  const validateEmail = (value: string) => {
+    if (value.includes('@')) {
+      return true;
+    }
+
+    return false;
+  };
```
anstsot, line 53: it would be good to add better validation and simplify the function: 
```suggestion
  const validateEmail = (value: string) => {
    return value.match(/.*\@.*\.\w{2,3}/g);
  };
```

src/components/PostDetails.tsx
```diff
@@ -1,96 +1,43 @@
...
-      // setComments([...comments, newComment]);
-      // works wrong if we wrap `addComment` with `useCallback`
-      // because it takes the `comments` cached during the first render
-      // not the actual ones
-    } catch (error) {
-      // we show an error message in case of any error
-      setError(true);
+      deleteComment(commentId);
+    } catch {
+      dispatch(loadComments(post.id));
```
anstsot, line 36: why do you need `loadComments` if delete throws an error? maybe it would be better to create additional action fore deleting?

anstsot (changes requested):
Good job, let's add some improvements and make your code better

src/components/PostDetails.tsx
savchuksavchuk: Hi, we need to remove comment immediately from the page. To avoid waiting for the response I decided delete comment from the store and then send request to the server. But if the some error occures, we need to reload actual list of comments from the server and put it into store.

polosanya (approved):

PR #233, opened 2023-03-07
solution w/redux-toolkit

eugenekh81:
[DEMO LINK](https://eugenekh81.github.io/react_redux-list-of-posts/) and add it to the PR description.


src/components/PostsList.tsx
```diff
@@ -1,56 +1,55 @@
...
-                  },
-                )}
-                onClick={() => {
-                  onPostSelected(post.id === selectedPostId ? null : post);
-                }}
-              >
-                {post.id === selectedPostId ? 'Close' : 'Open'}
-              </button>
-            </td>
+      <table className="table is-fullwidth is-striped is-hoverable is-narrow">
```
OleziO, line 15: Divide by a few lines.

src/components/NewCommentForm.tsx
```diff
@@ -60,7 +60,14 @@ export const NewCommentForm: React.FC<Props> = ({ onSubmit }) => {
     setSubmitting(true);
 
     // it is very easy to forget about `await` keyword
```
OleziO, line 62: Remove all comments from the code.

src/App.tsx
```diff
@@ -83,11 +71,7 @@ export const App: React.FC = () => {
                 )}
 
                 {author && loaded && !hasError && posts.length > 0 && (
```
OleziO, line 73: Replace with `posts.length`, it will be the same as `posts.length > 0`

OleziO (changes requested):
image
Add email validation.

image
Add string trim.

src/App.tsx
eugenekh81: it will not. If I do like you suggest, there will be a '0' in the place of posts.length. You are not the first mentor in Mate that recommends this approach, but this is not the correct one. You can verify it from your end, plus I am attaching a screenshot to show this. No changes made, leaving as is
Screenshot 2023-03-07 at 13 03 14

To avoid the appearance of '0' I'd need to wrap this condition into Boolean(posts.length), but in this case I don't think this makes my code shorter & easier to read.

src/components/PostsList.tsx
eugenekh81: validations added, email verification with regex added

src/App.tsx
eugenekh81: Thank you for the code review, changes made and deployed

vlad-demchuk (changes requested):
GJ!
I agree with you, `posts.length > 0` will be correct

vlad-demchuk (approved):

PR #242, opened 2023-03-30
solution

Deniskakaka:
https://deniskakaka.github.io/react_redux-list-of-posts/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,15 +15,15 @@ import {
   selectedPostAction,
 } from './features/posts/postsSlice';
 import { selecedUser } from './features/users/seletors';
-import { selectLoaded, selectorError } from './features/comments/selectors';
-import { selectorPosts } from './features/posts/selectors';
+import { selectorError } from './features/comments/selectors';
+import { selectorPosts, selectPostsLoad } from './features/posts/selectors';
 
 export const App: React.FC = () => {
   const dispatch = useAppDispatch();
   const author = useAppSelector(selecedUser);
   const hasError = useAppSelector(selectorError);
   const posts = useAppSelector(selectorPosts);
-  const loaded = useAppSelector(selectLoaded);
+  const loaded = useAppSelector(selectPostsLoad);
 
   const [selectedPost, setSelectedPost] = useState<Post | null>(null);
 
--- a/src/components/PostDetails.tsx
+++ b/src/components/PostDetails.tsx
@@ -80,7 +80,7 @@ export const PostDetails: React.FC<Props> = ({ post }) => {
           </p>
         )}
 
-        {!loaded && !hasError && comments.length && (
+        {!loaded && !hasError && comments.length ? (
           <>
             <p className="title is-4">Comments:</p>
 
@@ -112,7 +112,7 @@ export const PostDetails: React.FC<Props> = ({ post }) => {
               </article>
             ))}
           </>
-        )}
+        ) : <></>}
 
         {!loaded && !hasError && !visible && (
           <button
--- a/src/features/comments/commentsSlice.ts
+++ b/src/features/comments/commentsSlice.ts
@@ -75,7 +75,8 @@ const commentsSlider = createSlice({
     builder.addCase(
       deleteCommentAction.fulfilled,
       (state, action:PayloadAction<number>) => {
-        state.comments?.filter((el:Comment) => el.id !== action.payload);
+        state.comments = state.comments
+          ?.filter((el:Comment) => el.id !== action.payload);
       },
     );
     builder.addCase(deleteCommentAction.rejected, (state) => {
--- a/src/features/posts/selectors.ts
+++ b/src/features/posts/selectors.ts
@@ -1,3 +1,4 @@
 import { RootState } from '../../app/store';
 
 export const selectorPosts = (state:RootState) => state.posts.listPosts;
+export const selectPostsLoad = (state:RootState) => state.posts.loader;


src/App.tsx
```diff
@@ -8,52 +8,45 @@ import { PostsList } from './components/PostsList';
...
 
 export const App: React.FC = () => {
-  const [posts, setPosts] = useState<Post[]>([]);
-  const [loaded, setLoaded] = useState(false);
-  const [hasError, setError] = useState(false);
+  const dispatch = useAppDispatch();
+  const author = useAppSelector(state => state.users.selectUser);
+  const hasError = useAppSelector(state => state.posts.error);
+  const posts = useAppSelector(state => state.posts.listPosts);
+  const loaded = useAppSelector(state => state.posts.loader);
```
OlEzhA0, line 23: would be better to create some file like `selectors.ts`

and export 
```
export const selectAuthor = (state: RootState) => state.users.selectUser;
export const selectPostError = (state: RootState) => state.posts.error


usage:
const author = useAppSelector(selectAuthor);
const hasError = useAppSelector(selectPostError);
```

we need to do it, if path to some value change, you'll need to change it in only 1 place

src/components/PostDetails.tsx
```diff
@@ -57,38 +63,19 @@ export const PostDetails: React.FC<Props> = ({ post }) => {
...
-
-      // setComments([...comments, newComment]);
-      // works wrong if we wrap `addComment` with `useCallback`
-      // because it takes the `comments` cached during the first render
-      // not the actual ones
-    } catch (error) {
-      // we show an error message in case of any error
-      setError(true);
-    }
+    const newComment = await commentsApi.createComment({
```
OlEzhA0, line 66: add `try/catch`

src/features/comments/commentsSlice.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+const initialState: IPostState = {
+  comments: [],
+  loader: false,
+  error: false,
+};
+
+export const getCommentsAction = createAsyncThunk(
+  'comments/GET',
```
OlEzhA0, line 19: add string values to constants

src/features/comments/commentsSlice.ts
```diff
@@ -0,0 +1,92 @@
...
+  async (id:number) => {
+    const result = await commentsApi.getPostComments(id);
```
OlEzhA0, line 21: try catch

OlEzhA0 (changes requested):

src/features/comments/commentsSlice.ts
Deniskakaka: Why? if my editor handles all steps, including the error

src/App.tsx
OleziO: I recommend you resolve this comment, cause its a better solution.

src/App.tsx
```diff
@@ -8,52 +8,45 @@ import { PostsList } from './components/PostsList';
...
-    getUserPosts(userId)
-      .then(setPosts)
-      .catch(() => setError(true))
-      // We disable the spinner in any case
-      .finally(() => setLoaded(true));
+    dispatch(getPostsAction(userId));
   }
 
   useEffect(() => {
     // we clear the post when an author is changed
```
OleziO, line 32: Remove all comments from the code.

src/App.tsx
```diff
@@ -76,13 +69,13 @@ export const App: React.FC = () => {
                   </div>
                 )}
 
-                {author && loaded && !hasError && posts.length === 0 && (
+                {author && !loaded && !hasError && posts.length === 0 && (
                   <div className="notification is-warning" data-cy="NoPostsYet">
```
OleziO, line 73: Divide this line by a few lines.

OleziO (changes requested):
Fix the delete button.
image

src/App.tsx
Deniskakaka: it is all not my decision i chenge only redux

src/App.tsx
Deniskakaka: it is not my decition i chenge only redux

Deniskakaka:
why, i made this functional in react module, like this in task i need only add redux, this comments is not important because this task won't be in cv

src/components/PostDetails.tsx
```diff
@@ -104,23 +90,23 @@ export const PostDetails: React.FC<Props> = ({ post }) => {
...
 
-        {loaded && hasError && (
+        {!loaded && hasError && (
           <div className="notification is-danger" data-cy="CommentsError">
             Something went wrong
           </div>
         )}
 
-        {loaded && !hasError && comments.length === 0 && (
+        {!loaded && !hasError && comments.length === 0 && (
```
sTorba24, line 103:
```suggestion
        {!loaded && !hasError && !comments.length && (
```

src/components/PostDetails.tsx
```diff
@@ -104,23 +90,23 @@ export const PostDetails: React.FC<Props> = ({ post }) => {
...
           <p className="title is-4" data-cy="NoCommentsMessage">
             No comments yet
           </p>
         )}
 
-        {loaded && !hasError && comments.length > 0 && (
+        {!loaded && !hasError && comments.length > 0 && (
```
sTorba24, line 109:
```suggestion
        {!loaded && !hasError && !!comments.length && (
```

sTorba24 (changes requested):
You did good job! Fix test warnings and previous comments.
It's good practice to get used to writing clean code everywhere, not only for your CV
image

src/components/PostDetails.tsx
Deniskakaka: i need rewrite api? seriosly? it wrote not me, this api be in react module, and those it is ok there is not?

src/App.tsx
anstsot: it would be better to move these selectors

anstsot (changes requested):
Good work! I've found 2 improvements and please move the selectors from `App.tsx`.
1. The remove button isn't working. I've tried several times but the comment is still present
image

2. It would be great to add email validation 
image

Deniskakaka:
Learn the implemented App and the example and reimplement it with Redux having the next slice:

users and remove UsersContext;
author that is currently in the App;
posts having 3 props loaded, hasError and items;
selectedPost;
comments with loaded, hasError and items props (keep the visible in the PostDetails);
NewCommentForm should keep its state.

Deniskakaka:
я пренципиально не буду делать это, потому что этого нет в звдача и я это делал, мне надо дальше продвигаться а не повторять одино и тоже

anstsot (changes requested):
Good, you can leave the email field as it is but in the working example the delete button is working. Please implement it as on example: https://mate-academy.github.io/react_redux-list-of-posts/#user-1

src/components/PostDetails.tsx
```diff
@@ -112,7 +112,7 @@ export const PostDetails: React.FC<Props> = ({ post }) => {
               </article>
             ))}
           </>
-        )}
+        ) : <></>}
```
IvanFesenko, line 115: better will return null if nothing needs to render

IvanFesenko (approved):

PR #258, opened 2023-04-25
Solution

G0odvin:
[DEMO LINK](https://G0odvin.github.io/react_redux-list-of-posts/)


polosanya (changes requested):
Let's fix all eslint errors, so tests can check your work
image

qoa11a (approved):

PR #259, opened 2023-04-25
solved task

Serghii29:
[DEMO LINK](https://Serghii29.github.io/react_redux-list-of-posts/)


src/components/PostsList.tsx
```diff
@@ -1,56 +1,57 @@
...
+import { addPost, removePost } from '../features/selectedPost';
 
-type Props = {
-  posts: Post[],
-  selectedPostId?: number,
-  onPostSelected: (post: Post | null) => void,
-};
+export const PostsList: React.FC = () => {
+  const { items } = useAppSelector(state => state.posts);
+  const { item } = useAppSelector(state => state.selectedPost);
```
qoa11a, line 9: It's better to rename it so it's understandable whet these variables store.
```suggestion
  const { items: posts } = useAppSelector(state => state.posts);
  const { item: selectedPost } = useAppSelector(state => state.selectedPost);
```

PR #286, opened 2023-07-06
Ready

sergik94:
https://sergik94.github.io/react_redux-list-of-posts/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect } from 'react';
 import 'bulma/bulma.sass';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -8,52 +8,47 @@ import { PostsList } from './components/PostsList';
...
-  const [posts, setPosts] = useState<Post[]>([]);
-  const [loaded, setLoaded] = useState(false);
-  const [hasError, setError] = useState(false);
+  const { posts, loaded, hasError } = useAppSelector(state => state.posts);
 
-  const [author, setAuthor] = useState<User | null>(null);
-  const [selectedPost, setSelectedPost] = useState<Post | null>(null);
+  const author = useAppSelector(state => state.authors);
+  const dispatch = useAppDispatch();
+  const selectedPost = useAppSelector(state => state.selectedPost);
```
IvanFesenko, line 24: Recomendation: Create selector and use them
`const selectedPostSelector = (state: State) => state.selectedPost;`
You can reused it everywhere, and can be sure if something will change - you will change in one place - so everything will works

```
const selectedPost = useAppSelector(selectedPostSelector);
```

IvanFesenko (approved):
Well done

PR #287, opened 2023-07-06
React_redux-list-of-posts

ruslan-yarosh:
[DEMO LINK](https://ruslan-yarosh.github.io/react_redux-list-of-posts/)


src/features/users/usersSlice.ts
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable no-param-reassign */
+import { PayloadAction, createAsyncThunk, createSlice } from '@reduxjs/toolkit';
+
+import { User } from '../../types/User';
+import { getUsers } from '../../api/users';
+
+export interface UsersState {
+  users: User[];
+  expanded: boolean,
+  userError: boolean,
```
ivbutsykin, line 10:
```suggestion
  error: boolean,
```

I think you can use `error` name and if need change it on destructing level:
```
const { error: userError } = useAppSelector(...);
```

Also you can think about `hasError` to highlight boolean type

src/features/posts/postsSlice.ts
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable no-param-reassign */
+import { createAsyncThunk, createSlice } from '@reduxjs/toolkit';
+import { Post } from '../../types/Post';
+import { getUserPosts } from '../../api/posts';
+
+export interface PostsState {
+  posts: Post[];
+  loaded: boolean;
```
ivbutsykin, line 8:
```suggestion
  isLoading: boolean;
```

Such variant is more distributed, but up to you

src/features/comments/commentsSlice.ts
```diff
@@ -0,0 +1,89 @@
...
+          comment.id !== action.meta.arg
+        ));
+      })
+      .addCase(removeComment.rejected, (state) => {
+        state.hasError = true;
+      });
+
+    builder
+      .addCase(addComment.fulfilled, (state, action) => {
+        state.comments.push(action.payload);
```
ivbutsykin, line 81: Hm, don't remember if it allowed. Perhaps better to add in this way:
```
state.comments = [...state.comments, action,payload];
```

src/features/users/usersSlice.ts
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable no-param-reassign */
+import { PayloadAction, createAsyncThunk, createSlice } from '@reduxjs/toolkit';
+
+import { User } from '../../types/User';
+import { getUsers } from '../../api/users';
+
+export interface UsersState {
+  users: User[];
+  expanded: boolean,
```
ivbutsykin, line 9: Such things shouldn't be in store. Use `useState` instead

src/components/NewCommentForm.tsx
```diff
@@ -59,136 +61,148 @@ export const NewCommentForm: React.FC<Props> = ({ onSubmit }) => {
...
-            value={name}
-            onChange={handleChange}
-          />
-
-          <span className="icon is-small is-left">
-            <i className="fas fa-user" />
-          </span>
+    hasError ? (
+      <p>Error</p>
+    ) : (
```
ivbutsykin, line 80: To avoid nesting here you can write in such way:
```
if (hasError) {
  return <p>Error</p>
}

return (...);

ivbutsykin (changes requested):
Great job!
Like your decision with appearing post after click.

Only one critical comment from me: it's store `expanded` value in the store, try to fix it.

DarkMistyRoom (approved):

PR #292, opened 2023-07-12
Solution

yana-kotsulym:


src/app/store.ts
```diff
@@ -1,15 +1,38 @@
...
+} from '@reduxjs/toolkit';
 // eslint-disable-next-line import/no-cycle
-import counterReducer from '../features/counter/counterSlice';
+import usersReducer, { UsersState } from '../features/users/userSlice';
+import postsReducer, { PostsState } from '../features/posts/postsSlice';
+import commentsReducer,
+{ CommentsState } from '../features/comments/commentsSlice';
 
-export const store = configureStore({
+export type AppThunkDispatch = ThunkDispatch<RootState, any, AnyAction>;
```
loralevitska, line 15: Don't use any for types. Add the correct type here
image

loralevitska (changes requested):
Almost done) Let's improve
I can add empty comment. Please fix it with [string.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)
image

etojeDenys (approved):
lgtm

PR #298, opened 2023-07-19
Solution

andrii-kovalskyi02:
[DEMO LINK](https://andrii-kovalskyi02.github.io/react_redux-list-of-posts/)


loralevitska (changes requested):
Almost done) Let's improve

- fix adding empty comments, use [string.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)
image

polosanya (changes requested):
New comment form should close on post change
new-comment
image
Check how it works on a working example
https://mate-academy.github.io/react_redux-list-of-posts/

VitaliyBodnar (approved):
Good job!

PR #309, opened 2023-07-29
Task completed

roman-nikitenko:


src/app/store.ts
```diff
@@ -1,10 +1,16 @@
 import { configureStore, ThunkAction, Action } from '@reduxjs/toolkit';
 // eslint-disable-next-line import/no-cycle
 import counterReducer from '../features/counter/counterSlice';
+import usersSlice from '../features/usersSlice';
```
lerastarynets, line 4: The task requires to create a slice for 'author' and 'selectedPost', but it seems they are missing in your implementation. Make sure to create and import these slices.

src/app/store.ts
```diff
@@ -1,10 +1,16 @@
...
+import postsSlice from '../features/postsSlice';
+import commentsSlice from '../features/commentsSlice';
 
 export const store = configureStore({
   reducer: {
     counter: counterReducer,
```
lerastarynets, line 10: The 'counter' reducer is not needed according to the task requirement. You might want to remove this line if the counter feature is not used anymore.

src/features/postsSlice.ts
```diff
@@ -0,0 +1,68 @@
...
+  name: 'posts',
+  initialState,
+  reducers: {
+    setPosts: (state, action) => {
+      return {
+        ...state,
+        posts: action.payload,
+      };
+    },
+    selectedPost: (state, action) => {
```
lerastarynets, line 35: The reducer name 'selectedPost' could be more descriptive, for example 'setSelectedPost', to better indicate its function.

src/features/commentsSlice.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export const addComment = createAsyncThunk(
+  'commentCreate/fetch',
+  (newComment: NewCommentData) => {
+    return createComment(newComment);
+  },
+);
+
+export const deleteCom = createAsyncThunk(
+  'deleteComment/fetch',
```
lerastarynets, line 37: The asyncThunk name should follow the convention 'domain/event'. Here it should be 'comments/deleteComment'.

src/features/commentsSlice.ts
```diff
@@ -0,0 +1,102 @@
...
+  comments: Comment[];
+  loaded: boolean;
+  hasError: boolean;
+  visible: boolean;
+  error: string;
+};
+
+const initialState: CommentsState = {
+  comments: [],
+  loaded: true,
```
lerastarynets, line 19: The initial state for 'loaded' should be false. Only after successful fetching of comments, it should be updated to true.

src/features/commentsSlice.ts
```diff
@@ -0,0 +1,102 @@
...
+  visible: false,
+  error: '',
+};
+
+export const init = createAsyncThunk('comments/fetch', (postId: number) => {
+  return getPostComments(postId);
+});
+
+export const addComment = createAsyncThunk(
+  'commentCreate/fetch',
```
lerastarynets, line 30: The asyncThunk name should follow the convention 'domain/event'. Here it should be 'comments/commentCreate'.

src/features/usersSlice.ts
```diff
@@ -0,0 +1,54 @@
+import { createAsyncThunk, createSlice } from '@reduxjs/toolkit';
+import { User } from '../types/User';
+import { getUsers } from '../api/users';
+
+type UsersState = {
+  users: User[],
+  loaded: boolean,
+  hasError: string,
```
lerastarynets, line 8: The name `hasError` for a string variable can be misleading. It seems like it should be a boolean variable. Consider renaming it to `errorMessage` or `error`.

src/App.tsx
```diff
@@ -8,66 +8,56 @@ import { PostsList } from './components/PostsList';
...
                   </p>
                 )}
 
-                {author && !loaded && (
+                {selectedUser && !loaded && (
                   <Loader />
                 )}
 
-                {author && loaded && hasError && (
+                {selectedUser && loaded && hasError && (
```
lerastarynets, line 60: pls move all long ternary expressions to separate consts

src/App.tsx
```diff
@@ -76,18 +66,14 @@ export const App: React.FC = () => {
                   </div>
                 )}
 
-                {author && loaded && !hasError && posts.length === 0 && (
+                {selectedUser && loaded && !hasError && posts.length === 0 && (
```
lerastarynets, line 69:
```suggestion
                {selectedUser && loaded && !hasError && !posts.length && (
```
see [falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)

src/App.tsx
```diff
@@ -76,18 +66,14 @@ export const App: React.FC = () => {
...
                   </div>
                 )}
 
-                {author && loaded && !hasError && posts.length > 0 && (
-                  <PostsList
-                    posts={posts}
-                    selectedPostId={selectedPost?.id}
-                    onPostSelected={setSelectedPost}
-                  />
+                {selectedUser && loaded && !hasError && posts.length > 0 && (
```
lerastarynets, line 75:
```suggestion
                {selectedUser && loaded && !hasError && !!posts.length && (
```

src/components/PostDetails.tsx
```diff
@@ -1,105 +1,56 @@
...
-
-  useEffect(() => {
-    loadComments();
-  }, []);
+  }
 
-  useEffect(loadComments, [post.id]); // Wrong!
-  // effect can return only a function but not a Promise
-  */
+  useEffect(loadComments, [selectedPost?.id]);
```
lerastarynets, line 24: You are calling useEffect with dependency 'selectedPost?.id', however this will cause useEffect to run every time 'selectedPost?.id' changes. It would be better to check if 'selectedPost' has changed inside the 'loadComments' function and then call 'dispatch'.

src/components/PostDetails.tsx
```diff
@@ -1,105 +1,56 @@
...
-      // because it takes the `comments` cached during the first render
-      // not the actual ones
-    } catch (error) {
-      // we show an error message in case of any error
-      setError(true);
+        postId: selectedPost?.id,
+      };
+
+      dispatch(commentActions.addComment(newComment));
+      dispatch(commentActions.actions.setComments(newComment));
```
lerastarynets, line 36: You are dispatching an action to set a new comment two times in a row. You may want to check if that is intentional, as it could lead to unexpected state updates.

src/components/PostDetails.tsx
```diff
@@ -1,105 +1,56 @@
...
-    setComments(
-      currentComments => currentComments.filter(
-        comment => comment.id !== commentId,
-      ),
-    );
-
-    await commentsApi.deleteComment(commentId);
+  const deleteComment = (commentId: number) => {
+    dispatch(commentActions.deleteCom(commentId));
+    dispatch(commentActions.actions.delComment(commentId));
```
lerastarynets, line 42: You are dispatching an action to delete a comment two times in a row. You may want to check if that is intentional, as it could lead to unexpected state updates.

src/features/usersSlice.ts
```diff
@@ -0,0 +1,54 @@
...
+  selectedUser: User | null,
+};
+
+const initialState: UsersState = {
+  users: [],
+  loaded: false,
```
lerastarynets, line 14:
```suggestion
  isLoading: false,
```

lerastarynets (changes requested):
Nice start! Let's fix the solution according to the checklist

src/app/store.ts
roman-nikitenko: selectedPost should be a separate file ? correct ?
But the author is the user, right?

src/components/PostDetails.tsx
roman-nikitenko: That's how it works. Data is loaded only when the selected post changes. I checked

src/components/PostDetails.tsx
```diff
@@ -1,110 +1,62 @@
...
     <div className="content" data-cy="PostDetails">
       <div className="block">
         <h2 data-cy="PostTitle">
-          {`#${post.id}: ${post.title}`}
+          {`#${setSelectedPost?.id}: ${setSelectedPost?.title}`}
         </h2>
 
         <p data-cy="PostBody">
-          {post.body}
+          {setSelectedPost?.body}
```
Anastasiia-Svintsova, line 54: Use destructuring for objects for better readability

src/components/PostDetails.tsx
```diff
@@ -154,18 +106,20 @@ export const PostDetails: React.FC<Props> = ({ post }) => {
...
 
-        {loaded && !hasError && !visible && (
+        {!loaded && !hasError && !visible && (
           <button
             data-cy="WriteCommentButton"
             type="button"
             className="button is-link"
-            onClick={() => setVisible(true)}
+            onClick={() => {
+              dispatch(commentActions.actions.setVisible(true));
```
Anastasiia-Svintsova, line 115: Same about the destructuring. You can destructure the setVisible method from the commentActions.actions

src/features/authorSlice.ts
```diff
@@ -0,0 +1,63 @@
...
+export const init = createAsyncThunk('users/fetch', () => {
+  return getUsers();
+});
+
+const authorSlice = createSlice({
+  name: 'users',
+  initialState,
+  reducers: {
+    setUser: (state, action) => {
+      // eslint-disable-next-line no-param-reassign
```
Anastasiia-Svintsova, line 31: Disable this rule for the whole file to not repeat it several times

Anastasiia-Svintsova (changes requested):
- You should handle all API errors 

https://github.com/mate-academy/react_redux-list-of-posts/assets/99403198/0012cfda-4472-44c7-86d6-95d077ab8858

- Same. The user should see the error message after loading. Now it looks strange 

https://github.com/mate-academy/react_redux-list-of-posts/assets/99403198/2ea52f31-1c7e-4518-9003-76b720705020

src/components/PostDetails.tsx
roman-nikitenko: Good idea, thanks!

src/features/authorSlice.ts
roman-nikitenko: Thanks I didn't think of that

src/components/PostDetails.tsx
roman-nikitenko: This options doesn't work, because I need to dispatch my action for change.

src/components/PostDetails.tsx
roman-nikitenko: I have in this file the same simple names, for instance body, title. I can't do destructuring.

lerastarynets (approved):
You've got this, keep pushing forward!

PR #324, opened 2023-08-09
the task was done

pushkarskiyrodion:
[DEMO LINK](https://pushkarskiyrodion.github.io/react_redux-list-of-posts/)


src/features/posts/postsSlice.ts
```diff
@@ -0,0 +1,44 @@
...
+      state.loaded = true;
+      state.posts = action.payload;
+    });
+    builder.addCase(postsThunk.pending, (state) => {
+      state.loaded = false;
+    });
+    builder.addCase(postsThunk.rejected, (state) => {
+      state.loaded = true;
+      state.hasError = true;
+    });
```
maxim2310, line 39:
```suggestion
    builder.addCase(postsThunk.fulfilled, (state, action) => {
      state.loaded = true;
      state.posts = action.payload;
    });
    
    builder.addCase(postsThunk.pending, (state) => {
      state.loaded = false;
    });
    
    builder.addCase(postsThunk.rejected, (state) => {
      state.loaded = true;
      state.hasError = true;
    });
```

maxim2310 (changes requested):
Good work, but lets make some improve:
1. add validation to email field
2. make sure that response on delete comment request is ok and only then filter comment
3. and check all request, because now if i have error with requests application is falling

etojeDenys (approved):
lgtm

PR #328, opened 2023-08-14
Solution

Zibi95:


OleziO (changes requested):
Add string trim
image

src/components/PostsList.tsx
```diff
@@ -3,54 +3,52 @@ import React from 'react';
...
-            <td className="has-text-right is-vcentered">
-              <button
-                type="button"
-                data-cy="PostButton"
-                className={classNames(
-                  'button',
-                  'is-link',
-                  {
+        <tbody>
+          {posts.map((post) => (
```
volodymyr-soltys97, line 30: Use a destruction for `post`

src/components/PostsList.tsx
```diff
@@ -3,54 +3,52 @@ import React from 'react';
...
-          </tr>
-        ))}
-      </tbody>
-    </table>
-  </div>
-);
+                  })}
+                  onClick={() => {
+                    onPostSelected(post.id === selectedPostId ? null : post);
+                  }}
```
volodymyr-soltys97, line 43: Move this logic to the helper function and use it here

volodymyr-soltys97 (changes requested):
Good job 👍 
This is not in the task and the demo, so you need to delete it
image

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2023-11-09 at 15 12 20
Consider clearing redundant data when the author is changed

anastasiiavorobiova (changes requested):
Great progress here! Almost done!
Screenshot 2023-12-14 at 16 36 40
Consider clearing the state and closing the comments while the author is changed

Zibi95:
Already did that. I forgot to deploy. 😅

src/components/UserSelector.tsx
```diff
@@ -69,20 +67,24 @@ export const UserSelector: React.FC<Props> = ({
...
+          {users.map((user) => {
+            const { id, name } = user;
+
+            return (
+              <a
+                key={id}
+                href={`#user-${id}`}
+                onClick={() => {
+                  handleUserSelect(user);
+                }}
```
volodymyr-soltys97, line 79:
```suggestion
                onClick={() => handleUserSelect(user)}
```

volodymyr-soltys97 (approved):
Great work 🔥

PR #332, opened 2023-08-17
Develop

yura-enttsel:
[DEMO LINK](https://yura-enttsel.github.io/react_redux-list-of-posts/)

### DIFF
--- a/src/components/PostsList.tsx
+++ b/src/components/PostsList.tsx
@@ -1,56 +1,58 @@
 import classNames from 'classnames';
-import React from 'react';
 import { Post } from '../types/Post';
+import { useAppDispatch, useAppSelector } from '../app/hooks';
+import { clearSelectedPost, setSelectedPost } from '../features/selectedPost';
 
-type Props = {
-  posts: Post[],
-  selectedPostId?: number,
-  onPostSelected: (post: Post | null) => void,
-};
+export const PostsList = () => {
+  const dispatch = useAppDispatch();
+  const { posts } = useAppSelector(state => state.posts);
+  const { selectedPost } = useAppSelector(state => state.selectedPost);
+  const selectedPostId: number = selectedPost ? selectedPost.id : 0;
 
-export const PostsList: React.FC<Props> = ({
-  posts,
-  selectedPostId = 0,
-  onPostSelected,
-}) => (
-  <div data-cy="PostsList">
-    <p className="title">Posts:</p>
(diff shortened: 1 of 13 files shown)


maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)
1. your email field is not validate email
2. when i delete comment and have error while request you whatever filter your list

yura-enttsel:
> Great job overall! Just couple tiny things needs to be fixed before approval)
> 
> 1. your email field is not validate email
> 2. when i delete comment and have error while request you whatever filter your list


we delete the comment immediately so as not to make the user wait long for the actual deletion

maxim2310 (changes requested):
In case if you want to delete immediately you must return comment if request is fail

yura-enttsel:
chech pls this solution

src/components/UserSelector.tsx
```diff
@@ -1,34 +1,25 @@
...
-  value: selectedUser,
-  onChange,
-}) => {
-  // `users` are loaded from the API, so for the performance reasons
-  // we load them once in the `UsersContext` when the `App` is opened
-  // and now we can easily reuse the `UserSelector` in any form
-  const users = useContext(UserContext);
+export const UserSelector = () => {
+  const dispatch = useAppDispatch();
+  const { users } = useAppSelector(state => state.users);
```
volodymyr-soltys97, line 9: You are accessing 'users' and 'author' individually. Consider using destructuring to access these properties. For example, you can write const { users, author } = useAppSelector(state => state).

src/components/PostsList.tsx
```diff
@@ -1,56 +1,58 @@
...
+                  type="button"
+                  data-cy="PostButton"
+                  className={classNames(
+                    'button',
+                    'is-link',
+                    {
+                      'is-light': post.id !== selectedPostId,
+                    },
+                  )}
+                  onClick={handlePostSelection(post)}
```
volodymyr-soltys97, line 47: The handlePostSelection function is being invoked immediately instead of being passed as a callback. Replace 'handlePostSelection(post)' with '() => handlePostSelection(post)'.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #336, opened 2023-08-21
solution

Maks-Tsarenko:
[DEMO LINK](https://Maks-Tsarenko.github.io/react_redux-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect } from 'react';
 import 'bulma/bulma.sass';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -8,52 +8,38 @@ import { PostsList } from './components/PostsList';
 import { PostDetails } from './components/PostDetails';
 import { UserSelector } from './components/UserSelector';
 import { Loader } from './components/Loader';
-import { getUserPosts } from './api/posts';
-import { User } from './types/User';
-import { Post } from './types/Post';
-import { Counter } from './features/counter/Counter';
+// import { getUserPosts } from './api/posts';
```
loralevitska, line 11: Remove all comments. Try it write obvious code that explain all we need without comments)

loralevitska (changes requested):
Almost perfect) Let's improve for more practice
- fix adding empty comments. Use [string.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)
image

sTorba24 (approved):

PR #343, opened 2023-09-01
Solution by DK

DimaK123:
[DEMO LINK](https://DimaK123.github.io/react_redux-list-of-posts/)


src/components/PostsList.tsx
```diff
@@ -1,56 +1,55 @@
...
+                    'button',
+                    'is-link',
+                    {
+                      'is-light': post.id !== selectedPost?.id,
+                    },
+                  )}
+                  onClick={() => {
+                    dispatch(setSelectedPost(post.id === selectedPost?.id
+                      ? null : post));
+                  }}
```
IvanFesenko, line 44: Create separate handler and pass reference here instead of using anonumous function in JSX

IvanFesenko (approved):
Good job. Check my comment

PR #366, opened 2023-09-17
task solution

Bonum-taurus:
https://Bonum-taurus.github.io/react_redux-list-of-posts/


volodymyr-soltys97 (changes requested):
Do you need this block? It is not on the example of the working version
image

src/components/PostsList.tsx
```diff
@@ -1,56 +1,63 @@
...
+              <td data-cy="PostTitle">{post.title}</td>
+              <td className="has-text-right is-vcentered">
+                <button
+                  type="button"
+                  data-cy="PostButton"
+                  className={classNames(
+                    'button',
+                    'is-link',
+                    {
+                      'is-light': post.id !== selectedPostId,
```
volodymyr-soltys97, line 46: Create a helper function for this check and use them everywhere

src/components/PostsList.tsx
```diff
@@ -1,56 +1,63 @@
...
+                    },
+                  )}
+                  onClick={() => {
+                    dispatch(selectedPostActions
+                      .setPost(post.id === selectedPostId ? null : post));
+                  }}
```
volodymyr-soltys97, line 52: Consider will be better if you create a helper function for this logic and use them here

src/components/PostsList.tsx
```diff
@@ -1,56 +1,63 @@
...
-        ))}
-      </tbody>
-    </table>
-  </div>
-);
+        </thead>
+
+        <tbody>
+          {posts.map(post => (
+            <tr key={post.id} data-cy="Post">
```
volodymyr-soltys97, line 35: You can use a destructuring for `post`

volodymyr-soltys97 (changes requested):
Great work 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Great work 🔥

PR #370, opened 2023-09-20
Solution

RomaSheva1987:
[DEMO LINK](https://RomaSheva1987.github.io/react_redux-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -25,6 +25,8 @@ export const App: React.FC = () => {
   const clearPosts = () => dispatch(postsActions.clear);
   const setSelectedPost = () => dispatch(set(null));
 
+  const isPosts = author && loaded && !hasError;
+
   useEffect(() => {
     setSelectedPost();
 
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -83,11 +72,7 @@ export const App: React.FC = () => {
                 )}
 
                 {author && loaded && !hasError && posts.length > 0 && (
```
lerastarynets, line 74: pls consider moving long ternaries to separate consts with consistent names

lerastarynets, line 74:
```suggestion
                {author && loaded && !hasError && !!posts.length && (
```

src/app/store.ts
```diff
@@ -1,10 +1,18 @@
 import { configureStore, ThunkAction, Action } from '@reduxjs/toolkit';
 // eslint-disable-next-line import/no-cycle
```
lerastarynets, line 2: you shouldn't disable this eslint rule, pls refactor your code so that you won't have to disable it

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

sTorba24 (approved):

PR #379, opened 2023-09-28
Solution

outofs:
https://outofs.github.io/react_redux-list-of-posts/


maxyma-sanx (approved):
Good solution

ArtemTeslenko (approved):
Good job! Be more careful with files names, it can confuse your colleagues.

PR #406, opened 2023-10-30
solution

Reject-4444:
https://reject-4444.github.io/react_redux-list-of-posts/


VitaliyBondarenko1982 (approved):
Counter was only for example and could be removed.
But anyway looks good.

PR #411, opened 2023-11-02
List of posts

Orchibald:
[DEMO LINK](https://orchibald.github.io/react_redux-list-of-posts/)


src/features/posts.ts
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable no-param-reassign */
+import { createAsyncThunk, createSlice } from '@reduxjs/toolkit';
+import { Post } from '../types/Post';
+import { getUserPosts } from '../api/posts';
+
+export const init = createAsyncThunk('posts/fetch', (userId: number) => {
```
TarasHoliuk, line 6:
```suggestion
export const getPosts = createAsyncThunk('posts/fetch', (userId: number) => {
```
use `getPosts` or `getPostsByUserId` etc. init is not clear name

src/features/comments.ts
```diff
@@ -0,0 +1,67 @@
+/* eslint-disable no-param-reassign */
+import { PayloadAction, createAsyncThunk, createSlice } from '@reduxjs/toolkit';
+import { getPostComments } from '../api/comments';
+import { Comment } from '../types/Comment';
+
+export const initComments = createAsyncThunk('comments/fetch', (
```
TarasHoliuk, line 6: the same - getComments / loadComments, etc.

TarasHoliuk: or fetchComments - to be consistent with `fetchUsers`

TarasHoliuk (changes requested):
LGTM generally - let's rename a few variables only

VitaliyBondarenko1982 (approved):
Looks good

PR #420, opened 2023-11-08
1st version

Grechana:
https://Grechana.github.io/react_redux-list-of-posts/


maxim2310 (changes requested):
Good work, but lets make some improvement 

1. i must not be able to create comment with name with only spaces and same with massage
image
2. if i have error while deleting comment, it must be returned
3. same while creating comment must not be created if request failed

SanyaBratashchuk (changes requested):
If there is an error while deleting comment it must be returned. If you have any questions feel free to ask in chat

SanyaBratashchuk (changes requested):
Comment shouldn't be deleted if there is an error during deletion

SanyaBratashchuk (approved):

PR #424, opened 2023-11-14
Develop

NataliiaHen:
[DEMO LINK](https://NataliiaHen.github.io/react_redux-list-of-posts/)


anastasiiavorobiova (changes requested):
Great work! A user is able to post comments with spaces only, consider fixing it 
Screenshot 2023-11-14 at 14 57 21
Screenshot 2023-11-14 at 14 57 57
Comments should be cleared when author is changed

PR #440, opened 2023-11-25
reduxListOfPosts

NasyTheSlayer:
[DEMO LINK](https://NasyTheSlayer.github.io/react_redux-list-of-posts/)


loralevitska (changes requested):
Almost perfect) Let's improve

- use [string.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) to avoid adding empty comments
image

loralevitska (approved):
Good job!

PR #456, opened 2023-12-07
react_redux-list-of-posts

sofiiaTriasko:
[DEMO LINK] : https://sofiatryasko.github.io/react_redux-list-of-posts/

### DIFF
--- a/src/components/NewCommentForm.tsx
+++ b/src/components/NewCommentForm.tsx
@@ -1,13 +1,14 @@
 import classNames from 'classnames';
 import React, { useState } from 'react';
-import { CommentData } from '../types/Comment';
+import { addNewComment } from '../slices/commentsSlice';
+import { useAppDispatch, useAppSelector } from '../app/hooks';
 
-type Props = {
-  onSubmit: (data: CommentData) => Promise<void>;
-};
-
-export const NewCommentForm: React.FC<Props> = ({ onSubmit }) => {
+export const NewCommentForm: React.FC = () => {
   const [submitting, setSubmitting] = useState(false);
+  const selectedPost = useAppSelector(
+    (state) => state.selectedPost.selectedPost,
+  );
+  const dispatch = useAppDispatch();
 
   const [errors, setErrors] = useState({
     name: false,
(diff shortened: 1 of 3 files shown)


src/slices/usersSlice.ts
```diff
@@ -0,0 +1,48 @@
...
+    });
+
+    builder.addCase(usersThunk.rejected, (state) => {
+      state.loaded = false;
+      state.hasError = true;
+    });
+  },
+});
+
+export const { setUsers } = usersSlice.actions;
```
TarasHoliuk, line 47: you never use it:

image

so list of users is empty

src/slices/postsSlice.ts
```diff
@@ -0,0 +1,55 @@
+import { createSlice, PayloadAction, createAsyncThunk } from '@reduxjs/toolkit';
+import { Post } from '../types/Post';
+import { getPosts } from '../api/posts';
```
TarasHoliuk, line 3: you should get posts for selected user

src/App.tsx
```diff
@@ -9,23 +9,32 @@ import { PostDetails } from './components/PostDetails';
...
+  const posts = useAppSelector((state) => state.posts.items);
+
+  const dispatch = useAppDispatch();
 
   function loadUserPosts(userId: number) {
     setLoaded(false);
 
     getUserPosts(userId)
-      .then(setPosts)
+      .then(dispatch(setPosts))
```
TarasHoliuk, line 37: use `postsThunk` here like action

TarasHoliuk (changes requested):
Hey! Keep going - if you have any questions - feel free to ask them :)

You need to create async thunk for each request and use it as action. Create common actions, like this:

image

when you just want to somehow change local variable and don't fetch anything

lerastarynets (changes requested):
I can now add post whose fields consist of white spaces only, also name and email shouldn't be cleared after a post is added
Screenshot at Dec 13 10-22-59
on add post the ADD button should have loader inside, there is no need to show a loader on the whole section
Screenshot at Dec 13 10-23-31

sTorba24 (approved):
Nicely done!

PR #459, opened 2023-12-09
Develop

RodionSav:
https://rodionsav.github.io/react_redux-list-of-posts/


Andriy-Shymkiv (changes requested):
I can't select user on demo link. please fix it and request a review again

src/components/UserSelector.tsx
```diff
@@ -1,7 +1,9 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import classNames from 'classnames';
-import { UserContext } from './UsersContext';
+// import { UserContext } from './UsersContext';
```
loralevitska, line 3: Remove comment. Fix all cases

loralevitska (changes requested):
Almost done) 
- Try to fix all tests 
image
- Use [string triming](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) to avoid adding empty comment
image
- Close this bar if I change the user 
image

volodymyr-soltys97 (approved):
Well done 🔥

PR #461, opened 2023-12-11
done

victorgri:
https://victorgri.github.io/react_redux-list-of-posts/


src/App.tsx
```diff
@@ -76,18 +61,14 @@ export const App: React.FC = () => {
...
                   </div>
                 )}
 
-                {author && loaded && !hasError && posts.length > 0 && (
-                  <PostsList
-                    posts={posts}
-                    selectedPostId={selectedPost?.id}
-                    onPostSelected={setSelectedPost}
-                  />
+                {author && !loading && !error && posts.length > 0 && (
```
lerastarynets, line 70:
```suggestion
                {author && !loading && !error && !!posts.length && (
```
**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

src/App.tsx
```diff
@@ -76,18 +61,14 @@ export const App: React.FC = () => {
                   </div>
                 )}
 
-                {author && loaded && !hasError && posts.length === 0 && (
+                {author && !loading && !error && posts.length === 0 && (
```
lerastarynets, line 64: consider moving long ternaries to a separate const with consistent name and use the const here

src/app/store.ts
```diff
@@ -1,10 +1,18 @@
 import { configureStore, ThunkAction, Action } from '@reduxjs/toolkit';
 // eslint-disable-next-line import/no-cycle
```
lerastarynets, line 2: pls refactor your code so that you won't have to disable `import/no-cycle` rule

src/components/PostsList.tsx
```diff
@@ -1,56 +1,52 @@
...
+                  type="button"
+                  data-cy="PostButton"
+                  className={classNames('button', 'is-link', {
+                    'is-light': post.id !== selectedPostId,
+                  })}
+                  onClick={() => {
+                    dispatch(
+                      setSelectedPost(post.id === selectedPostId ? null : post),
+                    );
+                  }}
```
lerastarynets, line 41: Pls move the function inside `onClick` handler to a separate const and use as a callback here

lerastarynets (changes requested):
I can now add comment whose fields consist of white spaces only, pls fix
emptyPost

PR #462, opened 2023-12-11
Develop

kseniia-chepur:
[DEMO LINK](https://kseniia-chepur.github.io/react_redux-list-of-posts/)

Please note, that I removed useEffect from UserSelector component and used another approach for dropdown - the dropdown is closed upon user selection

### DIFF
--- a/src/components/NewCommentForm.tsx
+++ b/src/components/NewCommentForm.tsx
@@ -1,12 +1,12 @@
 import classNames from 'classnames';
 import React, { useState } from 'react';
-import { CommentData } from '../types/Comment';
+import { useAppDispatch, useAppSelector } from '../app/hooks';
+import { postComment } from '../features/comments';
 
-type Props = {
-  onSubmit: (data: CommentData) => Promise<void>;
-};
+export const NewCommentForm: React.FC = () => {
+  const { selectedPost } = useAppSelector(state => state.posts);
+  const dispatch = useAppDispatch();
 
-export const NewCommentForm: React.FC<Props> = ({ onSubmit }) => {
   const [submitting, setSubmitting] = useState(false);
   const [errors, setErrors] = useState({
     name: false,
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,59 +1,52 @@
...
+    dispatch(postsActions.setSelectedPost(null));
 
     if (author) {
-      loadUserPosts(author.id);
-    } else {
-      setPosts([]);
+      dispatch(fetchPosts(author.id));
     }
-  }, [author?.id]);
+  }, [author]);
```
anastasiiavorobiova, line 40: Screenshot 2023-12-11 at 14 06 13

anastasiiavorobiova (changes requested):
Great start! Keep doing: 
Screenshot 2023-12-11 at 14 02 38
The comments should be cleared when the user is changed 
Screenshot 2023-12-11 at 14 02 06
A user shouldn't be able to post empty comments (with spaces only) Also, the click on the delete comment button deletes a bunch of comments, instead of the clicked one

anastasiiavorobiova (changes requested):
Great work! The last things to do: 
Consider closing the list by clicking outside the list 
Screenshot 2023-12-11 at 16 57 17
Screenshot 2023-12-11 at 16 56 26
The form should remember the user name and mail after submission. The form should be cleared by clicking on reset button

kseniia-chepur:
I fixed  form submission (name & email are kept after submission), however, I cannot fixed comment related to closing the dropdown by clicking outside the list, in the original code it was implemented via useEffect which adds eventListener on click event and changes 'expanded' value to false. Due to unclear reason this useEffect doesn't work for me, I discussed the matter in technical chat and mentor after checking confirmed the same - it doesn't work for unknown reasons (probably, because of certain updates).
 Here is the link to the chat where the matter was discussed: [link](https://mate.academy/chats/8412/threads/1001893)

Esceype (approved):
Well done!

PR #469, opened 2023-12-24
Develop

reseneweb:
I skipped 1 test because it does not find the element with data-cy=UserSelector which is present which is present, all other tests passed.
https://reseneweb.github.io/react_redux-list-of-posts/


etojeDenys (approved):
lgtm

PR #476, opened 2023-12-29
react_redux-list-of-posts task solution

ivaZaiets:
Hello! This is my solution for this task. Check it out! ^^

```
PS:
Please help me with a failing test. In the demo, the loader is visible, 
and I have no idea why it can't find such an element. 
All other tests pass. Waiting for a solution!

And I also checked that this test doesn't pass in the default project, 
which is strange. 
Maybe the issue is not in my logic..?
```

[DEMO LINK](https://ivaZaiets.github.io/react_redux-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,51 +1,44 @@
-import React, { useEffect, useState } from 'react';
+import React, { useCallback, useEffect } from 'react';
+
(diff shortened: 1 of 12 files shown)


YaroslavYarynych (approved):
I love this solution, **good job**🤟🤟🤟

PR #479, opened 2024-01-02
Added solution

FurArt:
[Visit site](https://furart.github.io/react_redux-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,44 +10,56 @@ import { UserSelector } from './components/UserSelector';
 import { Loader } from './components/Loader';
 import { getUserPosts } from './api/posts';
 import { User } from './types/User';
-import { Post } from './types/Post';
-import { Counter } from './features/counter/Counter';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { getUsers } from './api/users';
+import { addUser } from './features/users/usersSlice';
+import {
+  setPosts,
+  setError,
+  setLoaded,
+} from './features/posts/postsSlice';
+import {
+  setSelectPost,
+} from './features/selectedpost/selectedPost';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -82,11 +94,12 @@ export const App: React.FC = () => {
                   </div>
                 )}
 
-                {author && loaded && !hasError && posts.length > 0 && (
+                {author
+                && loaded && !hasError && posts.length > 0 && (
```
lerastarynets, line 98:
```suggestion
                && loaded && !hasError && !!posts.length && (
```

lerastarynets, line 98: consider moving long ternaries to a separate const with consistent name and use the const here

src/app/store.ts
```diff
@@ -1,10 +1,18 @@
 import { configureStore, ThunkAction, Action } from '@reduxjs/toolkit';
 // eslint-disable-next-line import/no-cycle
```
lerastarynets, line 2: consider moving long ternaries to a separate const with consistent name and use the const here

src/components/NewCommentForm.tsx
```diff
@@ -3,7 +3,8 @@ import React, { useState } from 'react';
 import { CommentData } from '../types/Comment';
 
 type Props = {
-  onSubmit: (data: CommentData) => Promise<void>;
+  // onSubmit: (data: CommentData) => Promise<void>;
```
lerastarynets, line 6: pls remove redundant comments

lerastarynets (changes requested):
I can now add post whose fields consist of white spaces only, pls fix
emptyPost

lerastarynets (changes requested):
When I add a post with empty fields, the error is shown, however the empty post is still being added, also pls fix prev comments

anastasiiavorobiova (approved):
Great work!

PR #482, opened 2024-01-05
stage-1

no4kar:
[DEMO LINK](https://no4kar.github.io/react_redux-list-of-posts/)


src/features/users/usersSlice.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const usersSlice = createSlice({
+  name: 'users',
+  initialState,
+  reducers: {
+  },
+  extraReducers: (builder) => {
+    builder
+      .addCase(fetchUsers.pending, (state) => {
+        state.status = 'loading';// eslint-disable-line no-param-reassign
```
denys-danyliuk, line 28: disable this rule once for entire file instead of disabling it every time

denys-danyliuk (approved):
GJ!
Also you don't need to create a separate folder for each slice, you can just leave them in features folder

PR #497, opened 2024-01-17
add task solution

Softjey:
I'm aware of the linter warning about dispatchers, but I'll never change it, so I don't think it's worth spending resources tracking it.

Also, I added the files to the ignore list of the linter, because it argued that I was modifying the state, but in fact these functions work with a copy of the state.
/src/features/commentsSlice/
/src/features/postsSlice.ts

[DEMO LINK](https://softjey.github.io/react_redux-list-of-posts/)


src/App.tsx
```diff
@@ -8,48 +8,62 @@ import { PostsList } from './components/PostsList';
...
-    getUserPosts(userId)
-      .then(setPosts)
-      .catch(() => setError(true))
-      // We disable the spinner in any case
-      .finally(() => setLoaded(true));
-  }
+  useEffect(() => {
+    getUsers()
+      .then(usersFromServer => dispatch(usersActions.set(usersFromServer)));
+  }, []);
```
Andriy-Shymkiv, line 29: please add `dispatch` to deps array

src/App.tsx
```diff
@@ -8,48 +8,62 @@ import { PostsList } from './components/PostsList';
...
+    dispatch(selectedPostActions.remove());
 
     if (author) {
-      loadUserPosts(author.id);
+      dispatch(postsActions.init(author.id));
     } else {
-      setPosts([]);
+      dispatch(postsActions.clear());
     }
   }, [author]);
```
Andriy-Shymkiv, line 39: same here

src/App.tsx
```diff
@@ -8,48 +8,62 @@ import { PostsList } from './components/PostsList';
...
 
+  const showLoader = author && loading;
+  const showError = author && !loading && hasError;
+  const showNoPostsYet = author && !loading && !hasError && posts.length === 0;
+  const showPostList = author && !loading && !hasError && posts.length > 0;
+
+  const setAutor = useCallback(
+    (currAuthor: User) => dispatch(authorActions.set(currAuthor)),
+    [],
```
Andriy-Shymkiv, line 48: please do not ignore this eslint warning as it is very important

src/components/PostDetails.tsx
```diff
@@ -1,94 +1,40 @@
...
-    await commentsApi.deleteComment(commentId);
+    setVisible(false);
+    dispatch(commentsActions.loadComments(post.id));
+  }, [post.id]);
+
+  const addComment = useCallback(async (commentData: CommentData) => {
+    await dispatch(commentsActions.addComment({
+      ...commentData,
+      postId: post.id,
+    }));
```
Andriy-Shymkiv, line 33: no need to await here I suppose, you can just return
```suggestion
   return dispatch(commentsActions.addComment({
      ...commentData,
      postId: post.id,
    }));
```

src/features/commentsSlice/thunks.ts
```diff
@@ -0,0 +1,36 @@
...
+      throw new Error('Failed to add comment');
+    }
+  });
+
+const removeComment = createAsyncThunk('comments/remove',
+  async (commentId: number, { dispatch }) => {
+    try {
+      await dispatch(commentsActions.localeRemoveComment(commentId));
+
+      return await deleteComment(commentId);
```
Andriy-Shymkiv, line 30: please check the difference between `return Promise.resolve()` `Promise.resolve()` and just `return`
```suggestion
      return deleteComment(commentId);
```

src/features/usersSlice.ts
```diff
@@ -0,0 +1,12 @@
+import { PayloadAction, createSlice } from '@reduxjs/toolkit';
+import { User } from '../types/User';
+
+const usersSlice = createSlice({
+  name: 'users',
+  initialState: [] as User[],
```
Andriy-Shymkiv, line 6: do you really need this `as`?

Andriy-Shymkiv (changes requested):

src/features/usersSlice.ts
Softjey: Yes, because otherwise it's never[]

src/App.tsx
Softjey: In the redux documentation, it is written that [the dispatch function will not change if we do not change the store](https://react-redux.js.org/api/hooks#:~:text=The%20dispatch%20function%20reference,to%20do%20just%20that%3A). And one of the three main principles of redux is 'Single source of truth'. From the documentation: ['The global state of your application is stored in an object tree within a single store.'](https://redux.js.org/understanding/thinking-in-redux/three-principles).

So we just pass unnecessary dependencies to useEffect() and waste more resources, they are not important. By the same example, the function that we should insert, but we don't insert, because the linter understands that it is not needed is setSomeState().

src/components/PostDetails.tsx
Softjey: I need the function to return a Promise<void>

Softjey: Promise<void>*

Softjey: Promise void - github delete triangle arrows

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-01-18 at 15 28 52
Consider clearing comments in the redux when the user is changed 
Screenshot 2024-01-18 at 15 27 48
A user can create an empty comment. Consider fixing it

VitaliyBondarenko1982 (approved):
Looks good

PR #503, opened 2024-01-21
Develop

vitalii-fedusov:
[DEMO LINK](https://vitalii-fedusov.github.io/react_redux-list-of-posts/)


src/App.tsx
```diff
@@ -59,11 +32,11 @@ export const App: React.FC = () => {
                   </p>
                 )}
 
-                {author && !loaded && (
+                {author && loading && (
                   <Loader />
                 )}
 
-                {author && loaded && hasError && (
+                {author && !loading && error && (
```
lerastarynets, line 39: consider moving long ternaries to a separate const with consistent name and use the const here
**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

src/App.tsx
```diff
@@ -72,18 +45,14 @@ export const App: React.FC = () => {
                   </div>
                 )}
 
-                {author && loaded && !hasError && posts.length === 0 && (
+                {author && !loading && !error && posts.length === 0 && (
```
lerastarynets, line 48:
```suggestion
                {author && !loading && !error && !posts.length && (
```

src/App.tsx
```diff
@@ -72,18 +45,14 @@ export const App: React.FC = () => {
...
                   </div>
                 )}
 
-                {author && loaded && !hasError && posts.length > 0 && (
-                  <PostsList
-                    posts={posts}
-                    selectedPostId={selectedPost?.id}
-                    onPostSelected={setSelectedPost}
-                  />
+                {author && !loading && !error && posts.length > 0 && (
```
lerastarynets, line 54:
```suggestion
                {author && !loading && !error && !!posts.length && (
```

src/app/store.ts
```diff
@@ -1,10 +1,23 @@
 import { configureStore, ThunkAction, Action } from '@reduxjs/toolkit';
 // eslint-disable-next-line import/no-cycle
```
lerastarynets, line 2: pls refactor your code so that you won't have to disable `import/no-cycle` rule

lerastarynets (changes requested):
I can now add post whose fields consist of white spaces only, pls fix

emptyPost

also comments aren't deleted, when i click on cross button

src/app/store.ts
vitalii-fedusov: image
image

vitalii-fedusov: // eslint-disable-next-line import/no-cycle
so it was already given to me as example

VitaliyBondarenko1982 (approved):
Looks good.

PR #513, opened 2024-01-27
Add task solution.

MaksymBevskyi:
[DEMO LINK](https://MaksymBevskyi.github.io/react_redux-list-of-posts/)


src/features/comments.ts
```diff
@@ -0,0 +1,66 @@
...
+    },
+  },
+  extraReducers: (builder) => {
+    builder
+      .addCase(fetchComments.pending, (state) => {
+        state.loaded = false;
+      })
+      .addCase(fetchComments.fulfilled, (state, action) => {
+        state.comments = action.payload;
+        state.loaded = true;
```
TarasHoliuk, line 54:
```suggestion
        state.isLoading = true;
```
`loaded` means it's already done, i.e. it's already loaded. So in this case, `state.loaded` here should be `false`. But I'd prefer to keep value, but change name

src/features/comments.ts
```diff
@@ -0,0 +1,66 @@
...
+    removeComment: (state, action: PayloadAction<number>) => {
+      state.comments = state.comments
+        .filter(comment => comment.id !== action.payload);
+    },
+    clearComments: (state) => {
+      state.comments = [];
+      state.loaded = false;
+      state.hasError = false;
+    },
+    setError: (state) => {
```
TarasHoliuk, line 43: Do you use it? I guess you can delete it

TarasHoliuk (changes requested):
You haven't added all files to git. Please, push **all** files to make review possible

PR #536, opened 2024-02-29
Solution

Absolution26:
[DEMO LINK](https://Absolution26.github.io/react_redux-list-of-posts/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The user will not be able to add empty data, need to fix it
image
image

SanyaBratashchuk (approved):

PR #550, opened 2024-03-15
develop

RomanSylman:
[DEMO LINK](https://RomanSylman.github.io/react_redux-list-of-posts/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The fields cannot be an empty, you need to fix it
image

RomanSylman:
> Good job 👍 Let's improve your code The fields cannot be an empty, you need to fix it image
I don't see a problem either through the demo link or through the local server, could you record a video or specify exactly where and what you noticed.

anastasiiavorobiova (changes requested):
Great work! A user can post comments with spaces only 
Screenshot 2024-03-15 at 15 08 14
Consider clearing comments when a user is changed 
Screenshot 2024-03-15 at 15 06 20

PR #558, opened 2024-03-21
solution

Yehorf21:
[DEMO LINK](https://Yehorf21.github.io/react_redux-list-of-posts/)


src/components/UserSelector.tsx
```diff
@@ -1,34 +1,33 @@
...
+  const { users } = useAppSelector(state => state.users);
+  const { author } = useAppSelector(state => state.author);
   const [expanded, setExpanded] = useState(false);
 
+  useEffect(() => {
+    const fetchUsers = async () => {
+      const fetchedUsers = await getUsers();
+
+      dispatch(usersActions.set(fetchedUsers));
+    };
```
loralevitska, line 20: Create user out of useEffect

loralevitska (changes requested):
Keep up the good work! Some comments to fix:

- try to pass all tests 
image
- disable ability to add empty comment. Use [triming](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)
image

PR #559, opened 2024-03-22
add task solution

valeriaoleynik:
[DEMO LINK](https://valeriaoleynik.github.io/react_redux-list-of-posts/)

### DIFF
--- a/src/components/NewCommentForm.tsx
+++ b/src/components/NewCommentForm.tsx
@@ -47,20 +47,28 @@ export const NewCommentForm: React.FC<Props> = ({ onSubmit }) => {
   const handleSubmit = async (event: React.FormEvent) => {
     event.preventDefault();
 
+    const trimmedName = name.trim();
+    const trimmedEmail = email.trim();
+    const trimmedBody = body.trim();
+
     setErrors({
-      name: !name,
-      email: !email,
-      body: !body,
+      name: !trimmedName,
+      email: !trimmedEmail,
+      body: !trimmedBody,
     });
 
-    if (!name || !email || !body) {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -9,38 +9,40 @@ import { PostDetails } from './components/PostDetails';
...
+    dispatch(selectedPostActions.set(null));
 
     if (author) {
       loadUserPosts(author.id);
     } else {
-      setPosts([]);
+      dispatch(postsActions.set([]));
     }
-  }, [author]);
+  }, [author, dispatch]);
```
anastasiiavorobiova, line 45:
```suggestion
  }, [author, dispatch, loadUserPosts]);
```

anastasiiavorobiova (changes requested):
Great work! A user shouldn't be able to add a comment with spaces only 
Screenshot 2024-03-22 at 19 19 53
Also, consider clearing comments if the author is changed 
Screenshot 2024-03-22 at 19 20 32

src/App.tsx
valeriaoleynik: in working Demo  we can to add a comment with spaces only
I thought that I have to change task just with Redux or also i have to change logic of component - NewCommentForm?

anastasiiavorobiova (changes requested):
Great work! Almost done! Screenshot 2024-03-25 at 15 00 11
Consider clearing comments if the user is changed. Also, please, fix the form. A user shouldn't be able to add a comment with spaces only. So consider showing an error if the form field is empty. Feel free to ask for help in the chat, we will deal with everything

etojeDenys (approved):
lgtm

PR #562, opened 2024-03-27
solution

yuron-maker:
[DEMO LINK](https://yuron-maker.github.io/react_redux-list-of-posts/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect } from 'react';
 import 'bulma/bulma.sass';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -8,37 +8,24 @@ import { PostsList } from './components/PostsList';
...
+    dispatch(setSelectedPost(null));
 
     if (author) {
-      loadUserPosts(author.id);
+      dispatch(loadPosts(author.id));
     } else {
-      setPosts([]);
+      dispatch(setPosts([]));
     }
   }, [author]);
```
TarasHoliuk, line 30: do not ignore lint suggestions - it's better to do it:
image

src/App.tsx
```diff
@@ -66,18 +53,14 @@ export const App: React.FC = () => {
...
                 )}
 
-                {author && loaded && !hasError && posts.length > 0 && (
-                  <PostsList
-                    posts={posts}
-                    selectedPostId={selectedPost?.id}
-                    onPostSelected={setSelectedPost}
-                  />
+                {author && loaded && !hasError && items.length > 0 && (
+                  <PostsList posts={items} selectedPostId={selectedPost?.id} />
```
TarasHoliuk, line 63: Do you use `selectedPost` in App.tsx only to pass `id` as prop? In this case, you can just get it from Redux Selector in `PostList` and remove here

src/components/PostDetails.tsx
```diff
@@ -5,59 +5,33 @@ import { NewCommentForm } from './NewCommentForm';
...
-    } finally {
-      setLoaded(true);
-    }
-  };
-
   useEffect(() => {
-    loadComments();
-  }, []);
+    dispatch(loadComments(post.id));
+  }, [post.id]);
```
TarasHoliuk, line 28: same - add suggested by lint dependency

src/components/PostsList.tsx
```diff
@@ -3,52 +3,55 @@
...
 };
 
-export const PostsList: React.FC<Props> = ({
-  posts,
-  selectedPostId = 0,
-  onPostSelected,
-}) => (
-  <div data-cy="PostsList">
-    <p className="title">Posts:</p>
+export const PostsList: React.FC<Props> = ({ posts, selectedPostId = 0 }) => {
```
TarasHoliuk, line 14:
```suggestion
export const PostsList: React.FC<Props> = ({ posts, selectedPostId }) => {
```
default value is redundant here

src/components/PostDetails.tsx
```diff
@@ -66,25 +40,23 @@ export const PostDetails: React.FC<Props> = ({ post }) => {
         postId: post.id,
       });
 
-      setComments(currentComments => [...currentComments, newComment]);
+      dispatch(addComment(newComment));
```
TarasHoliuk, line 43: just add asyncThunc for it and handle error and success there

TarasHoliuk (changes requested):
GJ :+1: 

Added a few small suggestions to improve your code

src/App.tsx
yuron-maker: No, I don't only use id

anastasiiavorobiova (changes requested):
Great work! Consider clearing comments if the user is changed 
Screenshot 2024-03-29 at 16 41 28
Screenshot 2024-03-29 at 16 40 48
A user shouldn't be able to write an empty comment with spaces only

TarasHoliuk (approved):
Well done ✔️

PR #574, opened 2024-04-07
solved

DenisGurskiy:
https://DenisGurskiy.github.io/react_redux-list-of-posts/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The comment cannot be empty, you need to fix it
image

volodymyr-soltys97 (approved):
Nice work 🔥

PR #596, opened 2024-04-27
Develop

vladyslav-zakharov:
https://Diliradon.github.io/react_redux-list-of-posts/


l4st1m0za (changes requested):
Good job, but you need to change this detail:
User can input in name and email just blank space ` ` and it will work, but should not 
image

anastasiiavorobiova (changes requested):
Great work! A user can add an empty comment. Consider fixing it 
Screenshot 2024-04-28 at 15 25 05

anastasiiavorobiova (approved):
Awesome work!

PR #602, opened 2024-05-03
solution

VitaliiNez:
https://vitaliinez.github.io/react_redux-list-of-posts/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The comment cannot be only spaces, you need to fix it 
image

Add a correct type to input with email 
image

lerastarynets (approved):
Keep up the great work!

PR #621, opened 2024-05-29
Develop

Spec0s:
[redux api rtk query](https://spec0s.github.io/react_redux-list-of-posts/)


anastasiia-tilikina (changes requested):
Great job that you've managed to use RTK query 🔥 

but some of your tests wasn't successful let's try to fix it

I guess you won't be able to pass all tests because your approach
 But here are some problems we need to fix
 
 - your form should keep name and email after comment submitting
 - when you close one post and open another one - form should be hidden again
 - you should show error message if comments fetch wasn't successful

anastasiia-tilikina (changes requested):
well done!

but still i dont have error message if comments are failed to load

PR #628, opened 2024-06-05
task solution

mykhailonl:
[DEMO LINK](https://mykhailonl.github.io/react_redux-list-of-posts/)


vadiimvooo (changes requested):
Good job, but there is one bug remains:
For now it's possible to add comment with empty name, email and comment text, so please fix it.
image

src/components/PostDetails.tsx
```diff
@@ -5,31 +5,45 @@ import { NewCommentForm } from './NewCommentForm';
...
+    dispatch(setCommentsLoading(false));
+    dispatch(setCommentsError(false));
     setVisible(false);
 
     commentsApi
       .getPostComments(post.id)
-      .then(setComments) // save the loaded comments
-      .catch(() => setError(true)) // show an error when something went wrong
-      .finally(() => setLoaded(true)); // hide the spinner
+      .then(commentsList => dispatch(fetchComments(commentsList))) // save the loaded comments
```
TarasHoliuk, line 41: By the way, you could use createAsyncThunk to handle all promise states in reducer instead of dispatching a few different actions here

But all previous comments solved - approved ✔️

TarasHoliuk (approved):
Approved with one important comment - just FYI

PR #629, opened 2024-06-06
add task solution

ArsenSkandalenko:
[DEMO](https://allaserhiienko.github.io/react_redux-list-of-posts/)


src/app/store.ts
```diff
@@ -1,10 +1,18 @@
 import { configureStore, ThunkAction, Action } from '@reduxjs/toolkit';
-// eslint-disable-next-line import/no-cycle
-import counterReducer from '../features/counter/counterSlice';
```
TarasHoliuk, line 3: delete the reducer as well

src/features/author.ts
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable no-param-reassign */
```
TarasHoliuk, line 1: why are there 2 variants of slices - in store/reducers and here - in features?
Delete one

src/utils/axiosClient.ts
```diff
@@ -14,13 +14,17 @@ export const client = {
     return response.data;
   },
 
+  async post<T>(url: string, data: unknown) {
+
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   async post<T>(url: string, data: any) {
```
TarasHoliuk, line 20:
```suggestion
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  async post<T>(url: string, data: unknown) {
```

TarasHoliuk: same below

TarasHoliuk (changes requested):
fix all lint errors and tests

PR #648, opened 2024-07-01
react redux list of posts

msdreams:
[DEMO LINK](https://msdreams.github.io/react_redux-list-of-posts/)

### DIFF
--- a/src/features/comments.ts
+++ b/src/features/comments.ts
@@ -1,0 +1,63 @@
+/* eslint-disable no-param-reassign */
+
+import {
+  PayloadAction,
+  Slice,
+  createAsyncThunk,
+  createSlice,
+} from '@reduxjs/toolkit';
+import { Comment } from '../types/Comment';
+import { getPostComments } from '../api/comments';
+import * as commentsApi from '../api/comments';
+
+export type CommentsState = {
+  comments: Comment[];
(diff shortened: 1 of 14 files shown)


src/features/comments.ts
```diff
@@ -0,0 +1,63 @@
...
+  (postId: number) => {
+    return getPostComments(postId);
+  },
+);
+
+export const commentsSlice: Slice<CommentsState> = createSlice({
+  name: 'comments',
+  initialState,
+  reducers: {
+    set: (state, action: PayloadAction<Comment>) => {
```
TarasHoliuk, line 36: set doesn't describe (unlike in author slice) it - just `add` or `addComment`

src/features/comments.ts
```diff
@@ -0,0 +1,63 @@
...
+      state.comments.push(action.payload);
+    },
+    deleteCom: (state, action: PayloadAction<number>) => {
```
TarasHoliuk, line 39: do not use short versions - `deleteComment` is definitely better
Or just `delete` in case you don't want to repeat the slice name in each action name. But such repeating might be useful when you import it somewhere as separate action

TarasHoliuk (approved):
Looks good to me 👍 
Added 2 comments about naming just FYI

PR #651, opened 2024-07-05
Develop

AnastassiaH:
[DEMO LINK](https://AnastassiaH.github.io/react_redux-list-of-posts/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. You need to add this data-cy attribute for tests 
image

2. The comment cannot be only spaces 
image

3. Add a correct type to input of email 
image

volodymyr-soltys97 (changes requested):
Almost done!
Let's fix the failed test
The message that user is not selected should be shown only after the loader is gone
image

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-07-07 at 11 01 12
Consider clearing the previous user data if the user is changed 
Screenshot 2024-07-07 at 11 02 03

volodymyr-soltys97 (approved):
Great work 🔥

PR #653, opened 2024-07-07
add solution

creonluv:
[DEMO LINK](https://creonluv.github.io/react_redux-list-of-posts/)


src/features/counter/comments.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export const commentsSlice = createSlice({
+  name: 'comments',
+  initialState,
+  reducers: {
+    setComments: (state, action: PayloadAction<Comment[]>) => {
+      state.comments = action.payload;
+    },
+    addComments: (state, action) => {
```
TarasHoliuk, line 23: use async thunk (`createAsyncThunk`) to handle necessary promise states in extra reducers

src/app/store.ts
```diff
@@ -1,10 +1,20 @@
...
 import counterReducer from '../features/counter/counterSlice';
+import { usersReducer } from '../features/counter/users';
+import { authorReducer } from '../features/counter/author';
+import { postsReducer } from '../features/counter/posts';
+import { selectedPostReducer } from '../features/counter/selectedPost';
+import { commentsReducer } from '../features/counter/comments';
 
 export const store = configureStore({
   reducer: {
     counter: counterReducer,
```
TarasHoliuk, line 12: delete redundant counter slice
It was only an example

TarasHoliuk (changes requested):
You should use async thunk in this task

src/App.tsx
```diff
@@ -8,37 +8,42 @@ import { PostsList } from './components/PostsList';
...
 import { UserSelector } from './components/UserSelector';
 import { Loader } from './components/Loader';
-import { getUserPosts } from './api/posts';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { setAuthor } from './features/author';
 import { User } from './types/User';
+import { clearPosts } from './features/posts';
+
+// import { setLoading, setPosts, setError } from './features/posts';
+/* eslint-disable max-len */
```
volodymyr-soltys97, line 17: Remove all comments
```suggestion
```

src/features/comments.ts
```diff
@@ -0,0 +1,108 @@
...
+};
+
+export const commentsSlice = createSlice({
+  name: 'comments',
+  initialState,
+  reducers: {},
+  extraReducers: builder => {
+    builder
+      .addCase(loadComments.pending, state => {
+        /* eslint-disable-next-line no-param-reassign */
```
volodymyr-soltys97, line 51: Add this eslint rule only to the beginning of the file and remove it everywhere else

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The comment cannot be only spaces, you need to fix it 
image

Add a correct type to input of email 
image

src/App.tsx
```diff
@@ -8,37 +8,39 @@ import { PostsList } from './components/PostsList';
 import { PostDetails } from './components/PostDetails';
 import { UserSelector } from './components/UserSelector';
 import { Loader } from './components/Loader';
-import { getUserPosts } from './api/posts';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { setAuthor } from './features/author';
 import { User } from './types/User';
+import { clearPosts } from './features/posts';
+import { setSelectedPost as setSelectedPostAction } from './features/selectedPost';
```
TarasHoliuk, line 15:
```suggestion
import { setSelectedPost } from './features/selectedPost';
```
And lint error won't be a problem anymore
Generally - it was suggested to remove `// disable lint rule`. In such case you should try to fix the error - not just silently ignore it 🙂

src/App.tsx
```diff
@@ -8,37 +8,39 @@ import { PostsList } from './components/PostsList';
...
-    setSelectedPost(null);
+    dispatch(setSelectedPostAction(null));
 
     if (author) {
       loadUserPosts(author.id);
     } else {
-      setPosts([]);
+      dispatch(clearPosts());
     }
   }, [author]);
```
TarasHoliuk, line 45: just add dispatch here

image

src/features/author.ts
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable-next-line no-param-reassign */
```
TarasHoliuk, line 1: not next line - for the entire file (as in comments.ts)
And fix in other files

image

TarasHoliuk (changes requested):
Almost done! Let's fix linter errors

TarasHoliuk (approved):

PR #665, opened 2024-07-21
add task solution

wasik07:
[DEMO LINK](https://wasik07.github.io/react_redux-list-of-posts/)


src/features/comments.ts
```diff
@@ -0,0 +1,91 @@
...
+  visibleForm: false,
+  submitting: false,
+};
+
+export const initComments = createAsyncThunk(
+  'comments/fetch',
+  (postId: number) => commentsService.get(postId),
+);
+
+export const initCommentDeleted = createAsyncThunk(
```
TarasHoliuk, line 19: no need to name every action starting with init
It can be just `fetchComments`, `deleteComment`, `createComment`

Same in other slices

src/features/comments.ts
```diff
@@ -0,0 +1,91 @@
...
+  'comments/delete/fetch',
+  (commentId: number) => commentsService.delete(commentId),
+);
+
+export const initCommentCreat = createAsyncThunk(
```
TarasHoliuk, line 24: typo: creat -> create
I'd recommend using the Code Spell Checker extension for VSCode - very useful for me)

src/components/PostDetails.tsx
```diff
@@ -1,118 +1,66 @@
...
-        <h2 data-cy="PostTitle">{`#${post.id}: ${post.title}`}</h2>
+        <h2 data-cy="PostTitle">{`#${selectedPost?.id}: ${selectedPost?.title}`}</h2>
 
-        <p data-cy="PostBody">{post.body}</p>
+        <p data-cy="PostBody">{selectedPost?.body}</p>
       </div>
 
       <div className="block">
-        {!loaded && <Loader />}
+        {loaded && <Loader />}
```
TarasHoliuk, line 55: it seems misleading. If you show loader when this variable is `true` then it should be named `loading` (or better `isLoading`). Also, consider using only one variable instead of both for almost the same purpose (loading, submitting)

src/components/PostsList.tsx
```diff
@@ -1,54 +1,65 @@
...
+    ...state.comments,
+  }));
 
-type Props = {
-  posts: Post[];
-  selectedPostId?: number;
-  onPostSelected: (post: Post | null) => void;
-};
+  const getComments = (post: Post) => {
+    if (selectedPost?.id === post.id) {
```
TarasHoliuk, line 16: You can fetch comments inside PostDetails component and get rid of comparing id (just fetch comments for selectedPost)

src/components/UserSelector.tsx
```diff
@@ -1,62 +1,72 @@
...
+    ...state.posts,
+  }));
+
+  const selectUser = useCallback(
+    (newUser: User) => () => {
+      if (selectedUser?.id === newUser.id) {
+        return;
+      }
+
+      if (selectedPost) {
```
TarasHoliuk, line 23: same as my comment in PostList - do not fetch posts here in this function and this component (it should work with user only). You can just keep the original structure and fetch posts for user in useEffect in App.tsx

TarasHoliuk (changes requested):
GJ 👍 
Added a few refactoring suggestions

src/components/PostDetails.tsx
wasik07: I think they will conflict during simultaneous asynchronous actions. (Also, consider using only one variable instead of both for almost the same purpose (loading, submitting))

src/features/comments.ts
wasik07: I didn't find it)

VitaliyBondarenko1982 (approved):
Looks good

PR #695, opened 2024-08-22
Develop

GeorgeSavostikov:
[DEMO LINK](https://georgesavostikov.github.io/react_redux-list-of-posts/)


src/App.tsx
```diff
@@ -9,36 +9,36 @@ import { PostDetails } from './components/PostDetails';
...
+    dispatch(postsActions.setLoaded(false));
 
     getUserPosts(userId)
-      .then(setPosts)
-      .catch(() => setError(true))
-      // We disable the spinner in any case
-      .finally(() => setLoaded(true));
+      .then(serverPosts => dispatch(postsActions.setPosts(serverPosts)))
+      .catch(() => dispatch(postsActions.setError(true)))
+      .finally(() => dispatch(postsActions.setLoaded(true)));
```
TarasHoliuk, line 32: In this task, you should use `createAsyncThunk` to handle promise states (where you do it) in the reducer

src/app/store.ts
```diff
@@ -1,15 +1,24 @@
...
 import counterReducer from '../features/counter/counterSlice';
+import usersReducer from '../features/users/usersSlice';
+import authorReducer from '../features/author/authorSlice';
+import postsReducer from '../features/posts/postsSlice';
+import selectedPostReducer from '../features/selectedPost/selectedPostSlice';
+import commentsReducer from '../features/comments/commentsSlice';
 
 export const store = configureStore({
   reducer: {
     counter: counterReducer,
```
TarasHoliuk, line 11: delete counter - it was an example

TarasHoliuk (changes requested):
GJ 👍  But you should use async thunk in this task

volodymyr-soltys97 (changes requested):
Good job! 👍 
Let's improve your code
The comment cannot be only spaces, need to fix it 
image

Add a correct type to input of email 
image

volodymyr-soltys97 (approved):
Good job! 👍

PR #720, opened 2024-09-21
task solution

metrakonon:
[DEMO](https://metrakonon.github.io/react_redux-list-of-posts/)


src/store/authorSlice.ts
```diff
@@ -0,0 +1,23 @@
...
+const initialState: AuthorState = {
+  author: null,
+};
+
+export const authorSlice = createSlice({
+  name: 'author',
+  initialState,
+  reducers: {
+    setAuthor(state, { payload }: PayloadAction<User>) {
+      // eslint-disable-next-line no-param-reassign
```
undeletable, line 17: [Not critical]
If you see that some eslint rule conflicts with parts of your code and/or if some eslint rule doesn't make development more comfortable and failsafe, I'd avaive to consider switching this rule off on config level rather than disabling it for certain lines pr files.

src/App.tsx
```diff
@@ -6,39 +6,48 @@ import './App.scss';
...
+import { selectedPostSlice } from './store/selectedPostSlice';
 
 export const App: React.FC = () => {
-  const [posts, setPosts] = useState<Post[]>([]);
   const [loaded, setLoaded] = useState(false);
   const [hasError, setError] = useState(false);
 
-  const [author, setAuthor] = useState<User | null>(null);
-  const [selectedPost, setSelectedPost] = useState<Post | null>(null);
+  const posts = useAppSelector(state => state.posts.posts);
```
undeletable, line 23: [Not critical]
It makes sense to define selector callback outside of the components, along with reducers/slices. For example:
```
// reducer.ts
const postsSelector = state => state.posts.posts;
export { postsSelector };

// component
import { postsSelector } from "./reducer.ts";
...
const posts = useAppSelector(postsSelector);
```
With this approach, knowledge of store structure is isolated within reducer module. So, if you'd have to restructure the store, you'd only have to make the changes inside reducer module and not in the components which use data from it.

PR #741, opened 2024-10-03
Develop

arthurfujii:
[DEMO LINK](https://arthurfujii.github.io/react_redux-list-of-posts/)


pedro-ruas (approved):
Arthur, after analyzing the failing test and your code, I came to the conclusion that there is a problem within the test. Seems the test is trying to find an element that should not be there and fails because it isn't there xD
Because of this, I believe the issue isn't on your side and thus I am approving your task

PR #742, opened 2024-10-03
add task solution

vneholiuk:
[DEMO LINK](https://vneholiuk.github.io/react_redux-list-of-posts/)


src/slices/usersSlice.ts
```diff
@@ -0,0 +1,42 @@
...
+  async () => {
+    try {
+      const result = await getUsers();
+
+      return result;
+    } catch (error) {
+      throw error;
+    }
+  },
+);
```
etojeDenys, line 25:
```suggestion
export const fetchAllUsers = createAsyncThunk<User[], void>(
  'users/fetchAllUsers',
   getUsers
 );
```

src/slices/postsSlice.ts
```diff
@@ -0,0 +1,59 @@
...
+    try {
+      const respond = await getUserPosts(postId);
+
+      return respond;
+    } catch (error) {
+      throw new Error('Error');
+    }
+  },
+);
+
```
etojeDenys, line 30: you can simplify the same way each this function

etojeDenys (changes requested):
image
and it looks different in working example, pls don't hesitate to ask for some help in the chat

PR #744, opened 2024-10-04
add task solution

pawelnowicki87:
[DEMO LINK](https://pawelnowicki87.github.io/react_redux-list-of-posts/)


src/components/NewCommentForm.tsx
```diff
@@ -1,71 +1,64 @@
...
+        postId: selectedPost!.id,
+      }),
+    );
 
-    // and the spinner will disappear immediately
-    setSubmitting(false);
-    setValues(current => ({ ...current, body: '' }));
-    // We keep the entered name and email
+    dispatch(setValue({ field: 'body', value: '' }));
   };
```
Zibi95, line 62: I would keep state of from locally. You don't have to keep all the state inside store. You can still use useState or useReducer.

src/components/PostDetails.tsx
```diff
@@ -1,93 +1,25 @@
...
-    // not to make the user wait long for the actual deletion
-    // eslint-disable-next-line max-len
-    setComments(currentComments =>
-      currentComments.filter(comment => comment.id !== commentId),
-    );
-
-    await commentsApi.deleteComment(commentId);
-  };
+    setVisible(false);
+    dispatch(loadPostComments(selectedPost!.id));
```
Zibi95, line 21: Avoid using this ! symbol. Its non-null assertion operator in typescript. You can't never be sure that something will be null. If `selectedPost` will be `null` it will break whole app. Its safer to use this symbol ? or check if selectedPost exist before loading comments

src/components/PostItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+                    dispatch(
+                      selectPost(post.id === selectedPost?.id ? null : post),
+                    );
+                  }}
+                >
+                  {post.id === selectedPost?.id ? 'Close' : 'Open'}
+                </button>
+              </td>
+            </tr>
+          ))}
```
Zibi95, line 50: Use desctructure syntax whenever you can

src/components/PostItem.tsx
Zibi95, line 1: Why are there two PostsList components, and why is one of them in the PostItem file?

src/features/comments.ts
```diff
@@ -0,0 +1,117 @@
...
+  (data: Omit<Comment, 'id'>) => createComment(data),
+);
+
+export const removeComment = createAsyncThunk(
+  'comments/deleteComment',
+  async (commentId: number) => {
+    const delResponse = await deleteComment(commentId);
+
+    // eslint-disable-next-line no-console
+    console.log(delResponse);
```
Zibi95, line 35: clean up console.logs

Zibi95 (changes requested):
Hey. Good job! There are still some things to do. Clean up a bit and implement that PostItem component.

src/components/PostItem.tsx
pawelnowicki87: I need to use the whole obj 'post' in map method to get access to onClick method. In case of destructure id and title, i wont be able to get scope.

src/components/PostsList.tsx
```diff
@@ -21,32 +16,13 @@ export const PostsList: React.FC<Props> = ({}) => {
...
-                    dispatch(
-                      selectPost(post.id === selectedPost?.id ? null : post),
-                    );
-                  }}
-                >
-                  {post.id === selectedPost?.id ? 'Close' : 'Open'}
-                </button>
-              </td>
-            </tr>
+            <PostItem key={post.id} post={post} isSelected={selectedPost} />
```
Zibi95, line 25: I would expect boolean value in isSelected variable

src/components/PostItem.tsx
Zibi95: Oh okay. I didn;t notice that

src/components/PostDetails.tsx
```diff
@@ -1,165 +1,108 @@
...
+        {!comments.isCommetsLoading &&
+          comments.errorMessageOnCommentsLoading.length > 0 && (
+            <div className="notification is-danger" data-cy="CommentsError">
+              Something went wrong
+            </div>
+          )}
+
+        {!comments.isCommetsLoading &&
+          comments.errorMessageOnCommentsLoading.length === 0 &&
+          comments.comments.length === 0 && (
```
Zibi95, line 48: Here are also some multilevel boolean checks inside JSX

Zibi95 (changes requested):
Almost there! There are still some improvements that could be done.

natalia-klonowska (changes requested):
when deleting comment a loader is shown for the entire PostDetails component. this behavior is not user-friendly especially when you want to delete multiple comments cos it unnecessarily blocks the whole view

danon321 (approved):

PR #939, opened 2025-05-16
solution

fromjkeee16:
[DEMO LINK](https://fromjkeee16.github.io/react_redux-list-of-posts/)

### DIFF
--- a/src/features/commentsSlice.ts
+++ b/src/features/commentsSlice.ts
@@ -36,7 +36,7 @@ export const deletePostComment = createAsyncThunk(
 );
 
 export const commentsSlice = createSlice({
-  name: 'users',
+  name: 'comments',
   initialState,
   reducers: {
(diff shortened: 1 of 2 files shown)


alexander-ignatow (approved):
Good job overall, but there are a few issues:

Store Structure
The store structure is unnecessarily complex when it comes to slices.
For example, why are users and selectedUser separate, and similarly posts and selectedPost? Also, the naming is inconsistent.

Middleware Usage
It's great that you experimented with middleware, but you're not using it quite correctly.
The logic you've put in the middleware can (and should) be handled within the reducers instead.

Bug on Page Refresh
There's a small issue:

Select a user (the URL updates)

Refresh the page

The page reloads with the correct URL, but the selected user is no longer set

Overall, I don't think it's worth spending more time fixing this task. Let's move on to the other activities instead.

PR #1017, opened 2025-08-28
Task Solution

LitviVit:
[DEMO LINK](https://LitviVit.github.io/react_redux-list-of-posts/)


LitviVit:
@mateacademy-ai-mentor By the way, Redux Toolkit’s `createSlice` uses **Immer** under the hood. This means that you can either **mutate the draft state directly** or **return a new value**, and **Immer** will handle producing the correct next state. In the case of `selectedPostSlice` and `authorSlice`, returning the payload directly or `null` works correctly because Immer wraps the reducer, ensuring immutability under the hood. So keeping the current implementation is safe and fully compatible with Redux Toolkit’s best practices.

PR #1197, opened 2026-04-18
added task solution

AlexanderBarkar:


Anton-Kuchmasov (approved):
Good job, but in further tasks try to pass Linter check

PR #1223, opened 2026-05-19
reduxtool

DebHtm:
- Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
- Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
- This task does not have tests yet!
- Replace `<your_account>` with your Github username in the [DEMO LINK](https://DebHtm.github.io/react_redux-list-of-posts/) and add it to the PR description.


Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. the right block should also be displayed even if user isn't choosen
Image

Please, pay attention for Luke's comments.

etojeDenys (changes requested):
good job, but you should pass all the tests
Image

DebHtm:
Well, this error happens because I made the sidebar expandable, but according to the tests it should be closed by default. So the tests are failing because they detect it as opened, since that’s how it is in the template.

etojeDenys:
you can comment that test. also, this part should open after selecting post, but in your solution it open even if users aren't selected
image

DebHtm:
image
I can do that; it's just that my previous mentor showed me a screenshot where this panel was open even when no user was selected.

etojeDenys (approved):
good job

